PR #46, opened 2021-12-17
Port Travis CI to Github Actions 

primeroz:
* Ported all steps to a github action workflow
* On PR to main it will run the whole workflow excluding the release
  * Example RUN https://github.com/primeroz/kubecfg/actions/runs/1591802849
* On vX.X.X tag it will run the whole workglow including release
  * Create a Draft release named vX.X.X
  * Upload kubecfg artifacts  
  * Example RUN https://github.com/primeroz/kubecfg/actions/runs/1591832433

---
#### Differences from Travis version
* Golang is set to version 1.17
* Kind replaced minikube 
* There is an issue with k8s 1.20 where there is a CM created in all namespaces which causes the DELETE integration tests to fail. At the moment kind is locked to 1.19
* A Github Artifact is created to pass the binaries from the _build_ to the_release_ job. This will be available for 5 days in the non release actions run to test the binary if needed even before the PR is merged to master


.github/workflows/go.yml
```diff
@@ -1,25 +1,129 @@
...
+            git diff
+            echo 'Run "make generate" and add generated changes to this PR'
+            exit 1
+          fi
+          # Build kubecfg
+          make all
+
+      - name: Info Binary
+        run: |
+          #ldd ./kubecfg || otool -L ./kubecfg # What's the purpose of this ? it currently fail since we don't produce a dynamic executable
```
mkmik, line 55: on macos it is.

```console
$ otool -L ./kubecfg
./kubecfg:
	/usr/lib/libSystem.B.dylib (compatibility version 0.0.0, current version 0.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 0.0.0, current version 0.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 0.0.0, current version 0.0.0)
```

anyway it may be useful to show whether we indeed built a fully static executable or not.

.github/workflows/go.yml
```diff
@@ -1,25 +1,129 @@
...
+        run: |
+          make test
+          make vet
+
+      # Only run integration on linux matrix job
+      - name: Create k8s Kind Cluster
+        uses:  engineerd/setup-kind@v0.5.0
+        if: matrix.os == 'ubuntu-latest'
+        with:
+          version: "v0.11.1"
```
mkmik, line 69: whose version is this? kind's?

.github/workflows/go.yml
primeroz: ok, so what is the purpose ? to _succed_ if is a static file ? 

it was failing for me . mind if i re-enable it after this PR is merged ?

.github/workflows/go.yml
primeroz: yes, the image variable below define the k8s version

.github/workflows/go.yml
```diff
@@ -1,25 +1,129 @@
...
+          # NOTE - Starting with k8s 1.20 a CM is created in every namespace  called "kube-root-ca.crt" which breaks the "Simple Delete" integration test
+          image: "kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729"
+
+      - name: Integration Testing
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          go get github.com/onsi/ginkgo/ginkgo@$(go list -f '{{.Version}}' -m github.com/onsi/ginkgo)
```
mkmik, line 76:
```
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

```suggestion
          go install github.com/onsi/ginkgo/ginkgo@$(go list -f '{{.Version}}' -m github.com/onsi/ginkgo)
```

mkmik (commented):
looks good, just a couple of minor comments

.github/workflows/go.yml
primeroz: :+1:

.github/workflows/go.yml
mkmik: nah, let's just drop it

mkmik: I mean, if somebody wants to see that info, they can download the artifact and run the tool;

the only point of this info is to watch the mac info if you don't have a mac; it doesn't really matter I think.

.github/workflows/go.yml
primeroz: done

.github/workflows/go.yml
primeroz: :+1:

mkmik (approved):

PR #48, opened 2021-12-17
Run go mod tidy after renovate to cleanup go.sum and go.mod

primeroz:
This works in my local test to fix the issue we are getting on the prs from renovate but I have no way to test it will work in the full github workflow so i guess is worth a try , we can revert if it does not


primeroz:
this won't probably works since it runs after renovate. closing

PR #55, opened 2022-01-06
feat: Add helmTemplate builtin

primeroz:
This is an import of https://github.com/anguslees/kubecfg-1/tree/helm onto the new kubecfg repository

Add a new `helmTemplate` native function that downloads and expands
helm charts into jsonnet objects.  The new code links helm (v3)
libraries directly, and does not require nor use the external `helm`
command.

Fixes vmware-archive#267

Example usage:

```
local kubecfg = import "kubecfg.libsonnet";

local url = "https://charts.jetstack.io/charts/cert-manager-v1.5.3.tgz";
local cm = kubecfg.helmTemplate("cert-manager", namespace, url, {
  // Example values.yaml
  webhook: {replicaCount: 2},
});

// ... returns a jsonnet object with filename keys from the helm chart
// and expanded/parsed Kubernetes objects as values.

// The result can be manipulated and used just like any other jsonnet
// value.
cm + {
  "cert-manager/templates/webhook-deployment.yaml"+: {
    spec+: {
      template+: {
        spec+: {
          nodeSelector+: {"kubernetes.io/arch": "amd64"},
        },
      },
    },
  },
}
```

*Caveats:*

- Helm 'hooks' are not supported and ignored.
- Chart sort order is ignored, and the usual kubecfg sort mechanism is
  used.
- `chartURL` argument uses the kubecfg URL-based importer, but
  will reject relative URLs by default[1].
- HTTP_PROXY is obeyed, but there is no other cache.  The helm chart
  is re-downloaded on every invocation (for now).
- Probably some other things.

[1]: Relative URLs can be enabled using a new
`--allow-relative-helm-urls` flag.  URLs are interpreted relative to
$PWD currently, even when used by jsonnet from remote URLs.  _This will
change_.  TODO: Make this consistent with usual relative `import`
semantics, and enable by default.


primeroz:
@anguslees I ported your `kubecfg-1` `helm` branch to the new codebase rebased on current main. I did this mostly to speed up the work on this feature which i really really need ( and thanks for doing the real work on it! ) 

I only did 1 change here https://github.com/kubecfg/kubecfg/pull/55/commits/3efdcd12d85bcfc6d99e315c38448bbc04588c99 

```
for example using the kubernetes-dashboard chart the template `secret.yaml` return a list of secrets that wil render as

{
  'kubernetes-dashboard/templates/secret.yaml_0' : { apiVersion: v1, kind: secret, .... },
  'kubernetes-dashboard/templates/secret.yaml_1' : { apiVersion: v1, kind: secret, .... },
  'kubernetes-dashboard/templates/secret.yaml_2' : { apiVersion: v1, kind: secret, .... },
}

rather than

{
  'kubernetes-dashboard/templates/secret.yaml' : [
     { apiVersion: v1, kind: secret, .... },
     { apiVersion: v1, kind: secret, .... },
     { apiVersion: v1, kind: secret, .... },
   ]
}

to make it easier to further work on it with jsonnet.
```

I don't think is an issue since before when the result was a list you still needed to know the index of the resource in the templated you wanted to work on if you wanted to extend it but it makes it easier to work on the generated jsonnet with functions like `std.objectFields` and in general with the assumption that the output of the `helmTemplate` is a single dimension sequence of objects. But maybe there were other reasons not to do this ? 

I have been testing it and seems to work just fine , i was thinking to just add a few more tests in the `make jsonnettest` target

anguslees:
Awesome, thank you.

Re the output structure: I think we need a general list-to-object function, since I need to do something like this in numerous situations.  If we _had_ that generic function, then I think we should rely on that here, and our helm template function should return something true to the helm internals (ie: an object of arrays).

.. let me create some generic map/flatten functions and then see what we think about this question.

primeroz:
Sounds good, maybe we could also use it as part of the `ParseYaml` which at the moment returns a list and so i end up with 

```
local kubecfg = import 'kubecfg.libsonnet';

local yamlToObjects(str) = std.foldl(function(acc, i) acc {
  [i.apiVersion]+: {
    [i.kind]+: {
      [i.metadata.name]: i,
    },
  },
}, kubecfg.parseYaml(str), {});

{
  YAMLToObjects(str):: yamlToObjects(str),
}
```

Of course if you just want your own fork i will close this draft pr, this was really just to get the ball rolling and let me test the binary out of the latest codebase

anguslees:
This isn't being ignored.  I'm working on https://github.com/google/go-jsonnet/pull/584 so we can merge helmTemplate with support for relative helm import paths.  The result of https://github.com/kubecfg/kubecfg/pull/61 also impacts this PR.

primeroz:
closing in favour of https://github.com/kubecfg/kubecfg/pull/67

PR #57, opened 2022-01-10
Workaround sort instability without using replace directive

mkmik:
Using the `replace` directive is problematic, see https://github.com/kubecfg/kubecfg/pull/26#issuecomment-1008269344


pkg/kubecfg/show.go
```diff
@@ -25,7 +25,14 @@ import (
...
 	"github.com/Masterminds/sprig/v3"
-	yaml "gopkg.in/yaml.v2"
+
+	// yaml.v2 has a sort ordering bug that has been fixed in yaml.v3.
+	// However yaml.v3 hasn't been released yet. We tried to upgrade do yaml.v3 but it broke other tests.
+	// There is an open PR to backport the fix to v2 but it's stuck upstream for almost a year: https://github.com/go-yaml/yaml/pull/736
+	//
+	// Replace directives cause problems when installing with go install (see https://github.com/kubecfg/kubecfg/pull/26#issuecomment-1008269344)
+	// and also in other circumstances. We thus forked the go-yaml repo in:
+	"github.com/kubecfg/yaml/v2"
```
primeroz, line 35: don't we need to prefix this with `yaml` as it was before ? 

I ll just wait for CI to finish

primeroz (approved):

pkg/kubecfg/show.go
mkmik: no, there are two things here:

1. when you import a package, the name of the directory doesn't have to be the same as the name of the `package` directive. If the name is different it's good practice to actually put the name of the package in the import alias, and `goimports` (possibly also `gofmt`?) will do it for you.
2. the rules of go modules with version >1 state that you either have a `v2/` subdirectory in your actual repo or you just add `v2` to your `go.mod`'s `module` directive..

Goimports knows about rule (2) so it doesn't add an extra import alis.

PR #58, opened 2022-01-11
Fix regression in diff

### LINKED ISSUE #56: kubecfg diff behaviour is changed/broken?
Is there a breaking change in kubecfg-diff? Latest version show different output (not making much sense) than in v0.22.0.

mkmik:
Closes #56 

Due to an API change in the go-diff we can no longer compute per-line diffs, but instead we can render "word diffs", which in some cases may be more compact but are definitely harder to read (especially when not colorized):

---

**problem**:

go-diff new "by line" (buggy):

```
---
- live namespaces cerebro
+ config namespaces cerebro
  {
-   "apiVersion": "v1",
    "kind": "Namespace",
+ {
    "metadata": {
+ {
      "annotations": {},
+ {
      "labels": {
+ {
        "istio-injection": "enabled",
+ {
        "name": "cerebro"
+ {
      },
- {
+   "apiVersion": "v1",
- {
+   "apiVersion": "v1",
  {
- {
+   "apiVersion": "v1",
    "apiVersion": "v1",
+   }
+ }
```

go-diff native "pretty diff":

```
- live namespaces cerebro
+ config namespaces cerebro
  {
    "apiVersion": "v1",
    "kind": "Namespace",
    "metadata": {
      "- annotations": {},
-     "  labels": {
        "istio-injection": "enabled",
        "name": "cerebro+ ",
+       "some": "label  "
      },
      "name": "cerebro"
    }
  }
```

---

**solution**:

With the gotextdiff library we can get back the original output and also get a more compliant unified diff output (with proper context and spans).
```
--- live namespaces cerebro
+++ config namespaces cerebro
@@ -2,10 +2,10 @@
   "apiVersion": "v1",
   "kind": "Namespace",
   "metadata": {
-    "annotations": {},
     "labels": {
       "istio-injection": "enabled",
-      "name": "cerebro"
+      "name": "cerebro",
+      "some": "label"
     },
     "name": "cerebro"
   }
```

The ability of getting proper full line diffs is also important for kubecfg since we rely on that
to suppress Secret leakage with the `-omit-secrets` flag:

Now:

```
--- live secrets sec
+++ config secrets sec
@@ -1,6 +1,7 @@
+    baz: <omitted>
```

v0.22.0:

```
- live secrets sec
+ config secrets sec
+     baz: <omitted>
```


primeroz (approved):
LGTM

PR #61, opened 2022-01-15
Add isK8sObject, deepMap and fold jsonnet builtins

anguslees:
Add some helper functions to help manipulate jsonnet structures that
conform to kubecfg's usual "nested collections of Kubernetes objects"
result structure.

Specifically they handle:
- null (ignored)
- single Kubernetes object
- Kubernetes v1.List of Kubernetes objects
  (note other kind-specific list types are not supported)
- jsonnet array of any of these types (recursively)
- jsonnet object with values of any of these types (recursively)

Helper functions added in this commit:

`isK8sObject`:
returns true if this is a Kubernetes object.

`deepMap`:
map Kubernetes objects, preserving the original nested structure.

`fold`:
iterate over Kubernetes objects and accumulate a result.

`gvkName`:
useful function for `fold`. Accumulates a two-level nested object by
apiVersion.kind, then object name.

`gvkNsName`:
useful function for `fold`.  Accumulates a three-level nested object by
apiVersion.kind, object namespace (or `_`), then object name.

### DIFF
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -62,4 +62,75 @@
   // chart .tgz as an array of numbers (bytes).  `values` is a jsonnet
   // object that conforms to a schema defined by the chart.
   parseHelmChart:: std.native("parseHelmChart"),
+
+  // isK8sObject(o): Return true iff o is a Kubernetes object.
+  isK8sObject(o):: (
+    std.isObject(o) &&
+    std.objectHas(o, "apiVersion") &&
+    std.objectHas(o, "kind")
+  ),
+
+  // Private helper function for map/fold.
+  local isK8sList(o) = $.isK8sObject(o) && o.apiVersion == "v1" && o.kind == "List",
+
+  // deepMap(func, o): Apply the given function to each Kubernetes
+  // object in nested collection o, preserving the structure of o.
+  deepMap(func, o):: (
+    if isK8sList(o) then
+      o + {items: [func(item) for item in super.items]}
+    else if $.isK8sObject(o) then
+      func(o)
+    else if std.isObject(o) then
+      {[k]: $.deepMap(func, o[k]) for k in std.objectFields(o)}
+    else if std.isArray(o) then
+      [$.deepMap(func, elem) for elem in o]
+    else if o == null then
+      null
+    else
+      error ("o must be an object or array of k8s objects, found " + std.type(o))
+  ),
+
+  // fold(func, o, init): Apply the given function to each Kubernetes
+  // object in nested collection o, accumulating a result as we go.
+  // Function arg is invoked with arguments (accumulator, object).
+  fold(func, o, init):: (
+    if isK8sList(o) then
+      $.fold(func, o.items, init)
+    else if $.isK8sObject(o) then
+      func(init, o)
+    else if std.isObject(o) then
+      $.fold(func, [o[k] for k in std.objectFields(o)], init)
+    else if std.isArray(o) then
+      std.foldl(function (running, elem) $.fold(func, elem, running), o, init)
+    else if o == null then
+      init
+    else
+      error ("o must be an object or array of k8s objects, found " + std.type(o))
+  ),
+
+  layouts:: {
+    // gvkName(accum, o): Helper for 'fold'.  This accumulates a
+    // two-level collection of objects by 'apiVersion.kind'
+    // (GroupVersionKind) and then object 'name'.  NB: use gvkNsName
+    // if namespace is required for uniqueness.
+    gvkName(accum, o):: accum + {
+      [o.apiVersion + "." + o.kind]+: {
+        assert !(o.metadata.name in super),
+        [o.metadata.name]: o,
+      },
+    },
+
+    // gvkNsName(accum, o): Helper for 'fold'.  This accumulates a
+    // three-level collection of objects by 'apiVersion.kind'
+    // (GroupVersionKind), object 'namespace', and then object 'name'.
+    // Namespace is '_' for non-namespaced objects.
+    gvkNsName(accum, o):: accum + {
+      [o.apiVersion + "." + o.kind]+: {
+        [std.get(o.metadata, "namespace", default="_")]+: {
+          assert !(o.metadata.name in super),
+          [o.metadata.name]: o,
+        },
+      },
+    },
+  },
 }
--- a/testdata/kubecfg_test.jsonnet
+++ b/testdata/kubecfg_test.jsonnet
@@ -89,6 +89,71 @@ local testValue = [
 ];
 std.assertEqual(testValue, ["foo-mysql", "myns"]) &&
 
+std.assertEqual(kubecfg.isK8sObject("bogus"), false) &&
+
+std.assertEqual(kubecfg.isK8sObject({apiVersion: "v1", kind: "Pod"}), true) &&
+
+local obj(n) = {apiVersion: "example.com/v1alpha1", kind: "Test", name: n};
+local f(o) = o + {name+: "2"};
+local input = {
+  a: obj("a"),
+  b: [
+    {b1: obj("b1")},
+    obj("b2"),
+  ],
+  c: null,
+  d: {
+    apiVersion: "v1",
+    kind: "List",
+    extrakey: "foo",
+    items: [obj("d")],
+  },
+};
+local expected = {
+  a: obj("a2"),
+  b: [
+    {b1: obj("b12")},
+    obj("b22"),
+  ],
+  c: null,
+  d: {
+    apiVersion: "v1",
+    kind: "List",
+    extrakey: "foo",
+    items: [obj("d2")],
+  },
+};
+std.assertEqual(kubecfg.deepMap(f, input), expected) &&
+
+local obj(n) = {apiVersion: "example.com/v1alpha1", kind: "Test", name: n};
+local names(accum, o) = accum + [o.name];
+local input = {
+  a: obj("a"),
+  b: [
+    {b1: obj("b1")},
+    obj("b2"),
+  ],
+  c: null,
+  d: {
+    apiVersion: "v1",
+    kind: "List",
+    extrakey: "foo",
+    items: [obj("d")],
+  },
+};
+local expected = ["a", "b1", "b2", "d"];
+std.assertEqual(kubecfg.fold(names, input, []), expected) &&
+
+local obj(n) = {apiVersion: "example.com/v1alpha1", kind: "Test", metadata: {name: n}};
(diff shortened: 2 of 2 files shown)


anguslees:
Reviewers: We want to avoid function sprawl in kubecfg.libsonnet.  So please make suggestions if you can think of any other 'utility' functions that would be more generic/useful than these two (and a half).

One alternative is that we don't add these (since they aren't native functions and thus aren't _actually_ part of kubecfg).  I'm ok with moving these out into a 'unofficial' third-party-ecosystem jsonnet library, if we want to go that direction too.

anguslees:
@primeroz: As we discussed in #55, I think we needed some general 'helpers' for dealing with collections, more generally than just helm template output.  Let me know if you think this is reasonable to use:

```jsonnet
local h = kubecfg.helmTemplate("foo", "mynamespace", "kubernetes-dashboard-5.0.0.tgz", {});
// ^ this is now a 'correct' but rather inconvenient structure like
//    {"fileA.yaml": [obj], "fileB.yaml": [obj, obj]}
// In particular, you can't object-concatenate through those yaml-parsed arrays.

local byGvkName(accum, o) = accum + {[o.apiVersion + "." + o.kind]+: {[o.metadata.name]: o}};

local objs = kubecfg.k8sFold(byGvkName, h, {});
// ^ this is now a much nicer two-level nested-object structure like
//   {"some.apigroup/v1.Kind": {"objname": obj}}

// Example
obj + {
  "v1.Deployment"+: {   // <- apiVersion.kind
    "foo-kubernetes-dashboard"+: {  // <- metadata.name
      spec+: {
        strategy+: {
          rollingUpdate+: {
            maxSurge: 10,
          },
        },
      },
    },
  },
}
```

anguslees:
Oh, and I think I want to drop the `k8s` prefix and just have `kubecfg.deepMap` and `kubecfg.fold` - since I think the `kubecfg.` prefix is sufficient.  Thoughts?

primeroz:
Also i have not tested this yet (sorry bit busy ) , what is the output map for an input map like ` {"fileA.yaml": [obj], "fileB.yaml": [obj, obj]}` for `fileB.yaml` ?

anguslees:
> Is this correct ? since i don't see any changes to the helmTemplate function i think for single objects values we would not get a [] but just the object itself.

Oh right; if we are happy with this PR, I was going to propose moving `helmTemplate` back to just always returning an object-of-arrays - without treating single-element arrays as a special case.  I presumed this in my comment above, which was a bit eager - you're correct that the current `helmTemplate` PR will return `{"fileA.yaml": obj, "fileB.yaml": [obj, obj]}`.  The functions here will deal with both forms (and all other variations of nested collections).

> what is the output map for ...

See the test case for an example input/output: https://github.com/kubecfg/kubecfg/blob/905b9d1f24dfebc5e4f16a09216bbc2b6caae8a9/testdata/kubecfg_test.jsonnet#L85-L115

The 'structure' is preserved, and just the k8s objects themselves get converted by the map function (in the test case, I just append `2` to the `.name` property).

primeroz:
> Oh right; if we are happy with this PR, I was going to propose moving helmTemplate back to just always returning an object-of-arrays 

sounds good to me , as long as we provide helper jsonnet to convert to what people will probably want to use ?

anguslees:
> sounds good to me , as long as we provide helper jsonnet to convert to what people will probably want to use ?

I'm even less happy now that we're another step down the slippery slope :)  But I added `byGvkName`, which should do what folks want 95% of the time.

lib/kubecfg.libsonnet
```diff
@@ -56,4 +56,55 @@
...
+    else if o == null then
+      init
+    else
+      error ("o must be an object or array of k8s objects, found " + std.type(o))
+  ),
+
+  // byGvkName(accum, o): Helper for 'fold'.  This accumulates a
+  // two-level collection of objects by 'apiVersion.kind'
+  // (GroupVersionKind) and then object 'name'.
+  byGvkName(accum, o)::
```
mkmik, line 108: should we group such functions in a container like: ?

```
foldFormats: {
  byGvkName(accum, o):: ...
}
```

anguslees: I'm happy to do so if you want, but I think that just increases the typing boilerplate every time you want to use it.

mkmik: `kubecfg.layouts.gvkName`

`kubecfg.layouts.nsGvkName`

Yeah, I think paying a 6 chars price for a bit of self-documentation is worth it

lib/kubecfg.libsonnet
```diff
@@ -56,4 +56,55 @@
...
+    accum + {[o.apiVersion + "." + o.kind]+: {[o.metadata.name]: o}},
```
mkmik, line 109: Another thing: the name field is not mandatory. Objects could have a `genetatedName` field instead.

Perhaps we should handle that too

lib/kubecfg.libsonnet
anguslees: Ack, will change.   Just to be clear, an anticipated common case is now going to look like this:

```jsonnet
local objs = kubecfg.fold(kubecfg.layouts.gvkName, kubecfg.renderHelmChart(...), {});
```

One of the original objections to `fold(func, ..)` was the increased boilerplate from having to reference both fold and func (vs some hypothetical combined `toGvkName(input)` that did both in one go).  I'm surprised we're now eager to add extra characters here, which is why I'm resisting.

lib/kubecfg.libsonnet
anguslees: I agree, which is why this function only works 95% of the time.  What should we do for the generatedName case, since the object doesn't have a universal identifier in jsonnet?

For example, the following is a valid, useful kubernetes manifest.  It has zero way to separate the three jobs, other than by something contextual like position in the resulting array:

```jsonnet
local job = {apiVersion: "batch/v1", kind: "Job", metadata: {generateName: "myjob-"}};
[job, job, job]
```

Fwiw, here's the array index 'name' solution, that is the right thing for this specific case:

```jsonnet
local kubecfg = import "kubecfg.jsonnet";
local byPosition(accum, o) = accum + {["index" + std.toString(std.length(accum))]: o};

local objs = kubecfg.fold(byPosition, jobs, {});
objs + {
  index2+: {
     metadata+: {annotations+: {"foo": "I am an override for the 3rd job"}},
  },
}
```

Generates:
```yaml
---
apiVersion: batch/v1
kind: Job
metadata:
  generateName: myjob-
---
apiVersion: batch/v1
kind: Job
metadata:
  generateName: myjob-
---
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    foo: I am an override for the 3rd job
  generateName: myjob-
```

I think `fold` **is** the universal powerful function that you're looking for.  The accumulator function we pass to `fold` is by necessity a special case - that's why it's a _parameter_ that the user is meant to provide.

lib/kubecfg.libsonnet
mkmik: my main objection was not the number of chars per se, but whether it looks easy to understand to a casual user. Compare:

```
local objs = kubecfg.fold(kubecfg.layouts.gvkName, kubecfg.renderHelmChart(...), {});
```

with something like:

```
local objs = kubecfg.renderHelmChart(..., outputLayout=kubecfg.layouts.gvkName);
```

that said; we agreed to not go that route ^^^ and to keep the fold step explicit.

lib/kubecfg.libsonnet
mkmik: yeah; perhaps I'm biased; I rarely see multiple generatedNames with the same prefix in the same resource collection. I see it more used in order to generate new names every time they get applied.

I guess since we cannot cover 100% it's probably better to keep it simple.

mkmik, line 109: one last thing: 

the current byGvk function silently overwrites resources that have the same name:

```
{
  _demo:: [
    { kind: 'Config', apiVersion: 'v1', metadata: { name: 'foo' } },
    { kind: 'Config', apiVersion: 'v1', metadata: { name: 'foo' } },
  ],

  demo: kubecfg.fold(kubecfg.byGvkName, $._demo, {}),
```

produces:

```
{
   "demo": {
      "v1.Config": {
         "foo": {
            "apiVersion": "v1",
            "kind": "Config",
            "metadata": {
               "name": "foo"
            }
         }
      }
   }
}
```

perhaps we should error?

mkmik: something like:

```
byGvkName(accum, o):: accum { [o.apiVersion + '.' + o.kind]+: { assert !(o.metadata.name in super), [o.metadata.name]: o } },
```

mkmik: Or with some better error message:

```
  byGvkName(accum, o):: accum { [o.apiVersion + '.' + o.kind]+: local name = o.metadata.name; { assert !( name in super) : 'duplicate resource name %s' % name, [name]: o } },
```

mkmik: anyway we can improve this in another PR

anguslees: Oh!  This uncovers a good point.  It should probably be by gvk + _namespace_ + name.

anguslees: In particular, I think the above will show up a weakness in my existing renderHelmChart PR.  At the moment I add the namespace to _all_ jsonnet objects, regardless of whether they're _actually_ namespaced or not (k8s will ignore/drop it on apply, but it will look weird in a jsonnet function like proposed above).

mkmik: so how should we proceed here?

primeroz: I agree i never saw multiple objects with the same `generate-name` , the only times i saw the generate name is for jobs to trigger a recreate but is most time just one job

so i'd say we might be covering 99% of use cases already which is great

primeroz: in my personal experience i do have 2 functions similar to those , one with NS and one without it , and i use them depending on the situation

90% of the time the GVKName is enough since the `yaml` i am importing in jsonnet is all about a single namespace.

anguslees: > so how should we proceed here?

I don't care particularly :)  I just want something that y'all will approve so I can pop my PR stack and eventually merge the helm chart PR.

As mentioned previously, I think these 'layouts' functions are _necessarily_ opinionated and specialised (which is why I'd prefer not to have them in the kubecfg library at all).  If we add them (which I'm also ok with), then we shouldn't be surprised that they don't deal with all possible cases.

The PR now has a `gvkName` and `gvkNsName`.

primeroz (approved):
This LGTM . I will do more testing when the `ParseHelmChart` function is merged

mkmik (approved):

PR #65, opened 2022-02-04
Allow output filenames to have yml suffix

mark-rushakoff:
In our internal project, we have almost 5000 individual YAML files that
we generate. For historic reasons, we continue to name the files .yml
instead of .yaml.

By enabling kubecfg to write the files with a .yml suffix, we can
produce these files for our CI/CD pipeline measurably faster.

I used hyperfine to benchmark the current mode (my build of kubecfg, do
not specify --format, followed by find | read | mv), vs. the same build
of kubecfg with --format=yml, no call to find.

Benchmark #1: GENMODE=current make generate
  Time (mean ± σ):     13.384 s ±  0.459 s    [User: 79.086 s, System: 19.813 s]
  Range (min … max):   12.775 s … 14.280 s    10 runs

Benchmark #2: GENMODE=newkubecfg make generate
  Time (mean ± σ):     11.166 s ±  0.216 s    [User: 73.907 s, System: 9.905 s]
  Range (min … max):   10.858 s … 11.544 s    10 runs

Summary
  'GENMODE=newkubecfg make generate' ran
    1.20 ± 0.05 times faster than 'GENMODE=current make generate'

PR #66, opened 2022-02-04
Use --export-filename-extension instead of special casing -o yml

mkmik:
The more I think about it, the less I like using `-o yml` to represent yaml but with a different extension.
What if you want `*.YAML` extension or `.jsn` ?
More importantly I'm worried about forgetting to treat `yml` and `yaml` equivalently in the kubecfg internals.

The root issue is that `--export-filename-format` doesn't include the extension. This is in order to support outputting `.yaml` and `.json` files with `--export-dir` .

We should have added the extension as variable that can be referenced in the filename format.

Unfortunately we cannot add this and maintain backward compatibility.

This PR adds a new flag `-export-filename-extension` to override the extension.
Controlling the filename extension is quite a niche feature, so I think that adding a new flag is acceptable.

This PR effectively reverts #65

### DIFF
--- a/cmd/show.go
+++ b/cmd/show.go
@@ -16,22 +16,25 @@
 package cmd
 
 import (
-	"github.com/spf13/cobra"
+	"fmt"
 
 	"github.com/kubecfg/kubecfg/pkg/kubecfg"
+	"github.com/spf13/cobra"
 )
 
 const (
 	flagFormat               = "format"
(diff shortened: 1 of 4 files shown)


mark-rushakoff:
The rationale makes complete sense to me. Thanks.

primeroz (approved):

PR #67, opened 2022-02-05
Add parseHelmChart builtin

anguslees:
Add a new `parseHelmChart` native function that expands helm charts
into jsonnet objects.  The new code links helm (v3) libraries
directly, and does not require nor use the external `helm` command.

Note the helm chart is provided as an array of bytes (numbers 0-255) -
this builtin does not fetch the helm chart.  The expectation is that
this will be used with another mechanism like the jsonnet `importbin`
statement.

Fixes #267

Example usage:

```
local kubecfg = import "kubecfg.libsonnet";

local data = importbin "https://charts.jetstack.io/charts/cert-manager-v1.5.3.tgz";
local cm = kubecfg.parseHelmChart(data, "cert-manager", namespace, {
  // Example values.yaml
  webhook: {replicaCount: 2},
});

// ... returns a jsonnet object with filename keys from the helm chart
// and expanded/parsed Kubernetes objects as values.

// The result can be manipulated and used just like any other jsonnet
// value.
cm + {
  "cert-manager/templates/webhook-deployment.yaml": [o + {
    spec+: {
      template+: {
        spec+: {
          nodeSelector+: {"kubernetes.io/arch": "amd64"},
        },
      },
    },
  } for o in super["cert-manager/templates/webhook-deployment.yaml"]],
}
```

*Caveats:*

- Helm 'hooks' are not supported and ignored.
- Chart sort order is ignored, and the usual kubecfg sort mechanism is
  used.
- Probably some other things.


anguslees:
NB:
- The tests use the [jsonnet `importbin` statement](https://github.com/google/go-jsonnet/pull/584), and the feature (as implemented) is not very useful without it.  I strongly recommend we don't merge this until we can adopt a release of go-jsonnet that includes `importbin`.
- This PR supersedes/replaces #55 with a slightly different proposal.  We don't need that PR _and_ this PR.
- Has no direct relationship with #61, but those functions will make it much easier to work with the result structure from this builtin.

### Comparison with rest of jsonnet ecosystem

To keep some sanity for jsonnet library authors, we'd like to avoid conflicting builtin names with different semantics.  Survey is below, for posterity.  The name `parseHelmChart` is chosen to not conflict with any of these existing functions, and we could (eg) one day provide separate functions compatible with each of the following.

* [qbec] provides a `expandHelmTemplate` native function.

  Takes args (path, values, options)

  - `path` is helm chart as local tar.gz or directory.
  - `values` is standard helm values parameter object.
  - `options` is a custom object that basically captures each `helm template` command line flag.

* [tanka] provides a `helmTemplate` native function.

  Takes args (name, chart, opts)

  - `name` is chart release name
  - `chart` is local path to helm chart tgz
  - `opts` is for extra params, quite specific to the tanka implementation. Eg: The caller needs to set `calledFrom` to `std.thisFile` so relative chart paths work; `nameFormat` is a string in gotmpl format.


[qbec]: https://github.com/splunk/qbec/blob/bd1004ce37d976b7f500769b1d9e01a22d905860/site/content/reference/jsonnet-native-funcs.md#L8
[tanka]: https://github.com/grafana/tanka/blob/17da685bbff3baee75e529b13c19d8f047015395/pkg/helm/jsonnet.go#L34

mkmik:
@anguslees please remove the `vendor/` dir; we removed vendoring some time ago but I forgot to update the Makefile and the docs to not mention vendoring.

lib/kubecfg.libsonnet
```diff
@@ -56,4 +56,10 @@
   // to refer to submatches.  Regex is as implemented in golang regexp
   // package (python-ish).
   regexSubst:: std.native("regexSubst"),
+
+  // parseHelmChart(chartData, releaseName, namespace, values): Expand
```
mkmik, line 60: what about renaming it to `manifestHelm` ?

mkmik:
LGTM, except the name of the function

lib/kubecfg.libsonnet
mkmik: brain fart. yeah `manifestXX` is used to convert jsonnet into XX.

I understand that it makes sense to use `parse` to convey that it takes an external format and it "raises" it to jsonnet.

The thing that made confused me though is that we're not literally parsing a helm chart and presenting it as jsonnet; instead we're running a helm chart template expansion and parsing its result.

`parseRenderedHelmChyart` would be keyboardful

what about `renderHelmChart` or something similar?

CertainLach:
How do you plan to handle impurities of helm when evaluating charts?

I.e some charts use `randAlphaNum` template function, and while this does work, as helm is imperative, in jsonnet this breaks diff/patch with cluster, and causes unnecessary patching

In hayasaka (Analogous project, for which `jrsonnet` was initially written) i handle this by running templating twice, and require user to provide purification function (which should fix all impurities, by removing/changing fields to static values):
https://github.com/CertainLach/hayasaka/blob/trunk/src/helm/mod.rs#L143 (this repo is outdated, this tool is mostly used internally yet), example call can be found here: https://gist.github.com/CertainLach/334eac3058b1a41a7ec4383ffb04ac21#file-gitlab-jsonnet-L75-L79

Or do you plan on adding pure evaluation to helm itself?

Or haven't considered this, as almost all jsonnet to kubernetes deployment tools do? :D

mkmik:
@CertainLach it sounds like a neat trick! I certainly didn't consider this and thought I'd just handle "purifications" with normal jsonnet overlays once I notice the problem; but it surely makes a lot of sense to be able to detect this as early as possible.

anguslees:
> How do you plan to handle impurities of helm when evaluating charts?

As @mkmik mentioned, there's no plan here - other than 'buyer beware'.  The jsonnet user needs to use ad-hoc modifications to the input or results to avoid such impurities, where they affect the final output.

Hrm.  I think the best tool we have available is jsonnet, and I'd rather not introduce a different patch mechanism (eg json-patch).  Otoh, I like being able to highlight this issue automatically.  I don't think we can get both features without evaluating the jsonnet result again with a fresh JsonnetVM to ensure any jsonnet-level cache is flushed between runs - and at that point, the end-user may as well just do the same in a test that diffs output against a previous run.

(Related upstream helm feature request: https://github.com/helm/helm/issues/10689)

mkmik:
I assumed the suggested patch mechanism was just a jsonnet function applied before checking for diffs.

CertainLach:
> and I'd rather not introduce a different patch mechanism (eg json-patch).

In my case it is implemented in jsonnet, `purifier` function being called on every output object, before diff

> There's nothing that forces the URL contents to be 'immutable' and we leave it up to the jsonnet-user to do the right thing, or face the consequences

Well, this issue is possible to be solved too, as in nix package manager: requiring user to pass file hash, and print correct hash in case of mismatch:

https://github.com/NixOS/nixpkgs/blob/24e553ceabca4964ecf031c357ebc3376d4ad5c1/pkgs/os-specific/darwin/maloader/default.nix#L7-L12

Resulting value can be also cached by hash

mkmik:
yeah I like the idea of passing a hash to imports

anguslees:
> In my case it is implemented in jsonnet, purifier function being called on every output object, before diff

Oh! I'll read the code closer.  I only skimmed a few lines, saw references to json-patch, and jumped to an incorrect conclusion :(

CertainLach:
>  I think this is similar to the "diff against previous run" option above, only a checksum is smaller to store than the entire previous value.

Nothing is stored, in my case `helmTemplate` runs two helms in parallel, and then compares both outputs

It still vulnerable to `now` usage (Both helm runs may return same value), but most impurities comes from `random`, so this is fine enough

mkmik:
We cannot use hashes for helm output because the output is also a function of the "values" parameter you pass. In other words, hashes are only useful to handle impure 0-ary functions. Imports are 0-ary functions because the import url is not an expression. But we don't want to treat all prameters to parseHelmChart that way (only the chart package itself, hence `importbin`)

anguslees:
> We cannot use hashes for helm output because the output is also a function of the "values" parameter you pass

Yeah, I was imagining a hash of the final output of everything.  I agree this changes if you provide different input extvars (or similar).

---

So back to this PR, I think we're still down to the same choices:
1. Do nothing, and leave it up to people to do the right thing, or else understand the implications.  Or implement their own run-twice-and-compare check.
2. Evaluate helm output twice every time (with suitable cache invalidation) and compare, as a way of force-validating idempotency.
3. Evaluate and compare helm output against a known checksum.  This is a variant of (2) that compares runs from different times.

As mentioned earlier, this feels similar to URL imports.  I know from previous discussion that some folks see the ability for URL imports to change 'automatically' as a feature, not a bug. So I imagine that if we did (2) or (3) then we would need an option to turn it off (eg: maybe you _want_ to have today's date inserted into the helm result).

I'm still a fan of (1) (Do nothing).  I admit that's partly because it's the easiest to implement correctly :)  But also, it's super easy for folks to implement (2) or (3) themselves using a variety of techniques, if/when people want that extra assurance.  We can also add (2) or (3) as a later followup PR.

mkmik:
I think (2) and (3) are quite different. (2) protects from indeterministic output in the presence of constant inputs, while (3) protects about external side effects. (2) protects from instability here and now, (3) ensures long term stability (which some people want and other don't want).

Is there another way to protect from indeterministic evaluation? Setting a rng seed prior to evaluating helm?

There is a (1.5) option:

Run the evaluation twice, error if they two outputs don't match. Provide a boolean flag to ignore the error. This is simpler to implement and reason about in our first iteration and yet helps people who would be surprised by indeterministic helm charts

anguslees:
> Run the evaluation twice, error if they two outputs don't match. Provide a boolean flag to ignore the error. This is simpler to implement and reason about in our first iteration and yet helps people who would be surprised by indeterministic helm charts

Done.

Reviewers: I suggest you look at the commits separately, since this is now quite a large PR.

testdata/helm_test.jsonnet
```diff
@@ -0,0 +1,11 @@
+local kubecfg = import "kubecfg.libsonnet";
+
+// TODO: change to `importbin` with a sufficient version of jsonnet.
+local data = import "mysql-8.8.26.tgz.bin";
```
primeroz, line 4: How did you convert the TGZ to this `array of int` ?

anguslees: I used a version of jsonnet with `importbin` merged and printed the result of `importbin "mysql-8.8.26.tgz"`

mkmik: I hoped we could use `std.base64DecodeBytes` but it turns out it's implemented in jsonnet:

```
$ head -c100000 /dev/urandom | base64  >/tmp/foo.bin
$ jsonnet -e "local a=std.base64DecodeBytes(local b=importstr '/tmp/foo.bin'; b[:std.length(b)-1]); std.length(a)" 2>&1 | head
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc020240a88 stack=[0xc020240000, 0xc040240000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x11acf3a, 0x12ee240})
	/usr/local/Cellar/go/1.17.5/libexec/src/runtime/panic.go:1198 +0x71
runtime.newstack()
	/usr/local/Cellar/go/1.17.5/libexec/src/runtime/stack.go:1088 +0x5ac
runtime.morestack()
```

mkmik: one thing we could do while waiting for a release of go-jsonnet that includes support for `importbin` is to extend our custom importer to support something like:

```import 'binary-bytes:<url>'```

Comment (primeroz) [reply testdata/helm_test.jsonnet:4]: > one thing we could do while waiting for a release of go-jsonnet that includes support for importbin is to extend our custom importer to support something like:
> 
> import 'binary-bytes:<url>'

this sounds good to me , it would be backward compatbile when `importbin` is merged and we could add a deprecation message at that point

Comment (mkmik) [reply testdata/helm_test.jsonnet:4]: in the meantime, this python oneliner can be used to convert a binary file into a jsonnet compatible array of integers:
```
python -c 'print([ord(i) for i in open("/tmp/foo.tgz").read()])
```

anguslees: > is to extend our custom importer to support something like [...]

Fwiw, I don't think we should write new code to workaround importbin.  If we really want importbin _now_, then we can just pull a pre-release sha of go-jsonnet into kubecfg.

mkmik: there are not many other changes since 0.18.0 so this option appears to be doable; how can we make renovatebot not try to keep the go-jsonnet dep up to date to every latest sha though?

primeroz: this sounds good to me.

I see is already imported in the `go-jsonnet` while it is still open on the `jsonnet` one with an open comment.

PR #68, opened 2022-02-07
Remove go mod vendor from Makefile

mkmik:
I removed vendoring in #29 but forgot this command in the makefile.

Users who were used to run `make vendor` end up putting stuff back to `vendor/`, see #67


primeroz:
should we remove the docs from https://github.com/kubecfg/kubecfg/blob/main/CONTRIBUTING.md#managing-vendor ?

primeroz (approved):

PR #70, opened 2022-02-09
Use go1.16 FS embed

mkmik:
Massively simplify embedding `lib/*.*sonnet` files by leveraging
go 1.16 native embed.FS feature


anguslees (approved):
wow. ❤️

PR #82, opened 2022-03-08
Add provenance annotations

mkmik:
Kubecfg walks the input jsonnet tree and finds k8s objects scattered around in the jsonnet tree. Users are free to group objects the way it makes sense for them. This greatly helps composing modular systems of resources and also allows a meaningful overloading in different environment.

However, `kubecfg show` flattens out all that into a flat stream of k8s resources.

Sometimes you want to be able to "walk things backward" from the the final rendered manifest and I found that knowing where in the input tree (file + path) could be very helpful:

Demo:

```console
$ kubecfg show ./examples/guestbook.jsonnet --show-provenance | head
---
apiVersion: apps/v1beta2
kind: Deployment
metadata:
  annotations:
    kubecfg.github.com/provenance-file: ./examples/guestbook.jsonnet
    kubecfg.github.com/provenance-path: $.slave.deploy
  labels:
    name: redis-slave
  name: redis-slave
```

### DIFF
--- a/utils/acquire.go
+++ b/utils/acquire.go
@@ -33,8 +33,8 @@ import (
 )
 
 const (
-	AnnotationProvenanceFile = "kubecfg.dev/provenance-file"
-	AnnotationProvenancePath = "kubecfg.dev/provenance-path"
+	AnnotationProvenanceFile = "kubecfg.github.com/provenance-file"
+	AnnotationProvenancePath = "kubecfg.github.com/provenance-path"
 )
 
 type readOptions struct {


utils/acquire.go
```diff
@@ -32,10 +32,32 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
+const (
+	AnnotationProvenanceFile = "kubecfg.dev/provenance-file"
+	AnnotationProvenancePath = "kubecfg.dev/provenance-path"
```
anguslees, line 37: Should we talk about this namespace?  I think these are different to what we use elsewhere in kubecfg.  Should we be consistent, then (maybe) switch them all, or have different values with different prefixes depending on when the feature was added?  (I could be convinced either way)

anguslees (commented):
:heart:

utils/acquire.go
mkmik: yeah I'd like to switch the old ones to this. I just bought the domain so we have something that's affiliated only with this project

mkmik: an alternative domain name would be `kubecfg.github.com`, in case we don't want to fiddle with domain names (and possibly simplify future maintainership transferals); I quite like a simple `kubecfg.dev` home for the project fwiw.

mkmik: @anguslees how should we proceed?

primeroz: kubecfg.github.com is my favourite for what

mkmik: ok, let's go with `kubecfg.github.com`

primeroz (approved):

PR #90, opened 2022-04-01
Cache openapi schema in discovery client

anguslees:
Cache `OpenAPISchema()` results in our cached discovery client too.

kubecfg does several schema-driven operations and caching this
immutable information gives a huge speed boost.


mkmik (approved):
Thanks

PR #110, opened 2022-07-12
Implement 'binary://' hack in importer

mkmik:
This PR implements a hack to work around the lack of availability of `importbin`.

`importbin` has been merged in `go-jsonnet` (see https://github.com/google/go-jsonnet/pull/584) but:

* a new release hasn't been cut and it's not clear when will that happen
* the equivalent PR in the C++ repo hasn't been merged yet and it's not clear whether will that ever happen
* `kubecfg` could just depend on an unreleased version of `go-jsonnet` but end users would have trouble running `jsonnetfmt` and similar tooling when maintaining jsonnet sources.

This PR implements a hack abusing the import hooks:

```
import 'binary://someurl'
```

will produce an array of integers like `importbin` would.

Tested with:

```
{
  apiVersion: 'v1',
  kind: 'ConfigMap',

  data: {
        foo:: std.base64(import 'binary:///tmp/demo.bin'),
        md: std.md5(self.foo+'\n'),
  },
}
```

```console
$ kubecfg show demo.jsonnet
---
apiVersion: v1
data:
  md: e34aa50a97c327fbd5d1dcc427b9914d
kind: ConfigMap
$ md5sum <(base64 /tmp/ingress-nginx-4.1.4.tgz )
e34aa50a97c327fbd5d1dcc427b9914d  /dev/fd/11
```

### DIFF
--- a/utils/importer.go
+++ b/utils/importer.go
@@ -30,6 +30,7 @@ MakeUniversalImporter creates an importer that handles resolving imports from th
 In addition to the standard importer, supports:
   - URLs in import statements
   - URLs in library search paths
+  - importing binary files (for local files and URLs)
 
 A real-world example:
   - You have https://raw.githubusercontent.com/ksonnet/ksonnet-lib/master in your search URLs.
@@ -76,6 +77,12 @@ type universalImporter struct {
 func (importer *universalImporter) Import(importedFrom, importedPath string) (jsonnet.Contents, string, error) {
 	log.Debugf("Importing %q from %q", importedPath, importedFrom)
 
+	binary := false
+	if strings.HasPrefix(importedPath, "binary://") {
+		binary = true
+		importedPath = strings.TrimPrefix(importedPath, "binary://")
+	}
+
 	candidateURLs, err := importer.expandImportToCandidateURLs(importedFrom, importedPath)
 	if err != nil {
 		return jsonnet.Contents{}, "", fmt.Errorf("Could not get candidate URLs for when importing %s (imported from %s): %v", importedPath, importedFrom, err)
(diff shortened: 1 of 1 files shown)


primeroz (approved):

utils/importer.go
```diff
@@ -64,18 +65,29 @@ func MakeUniversalImporter(searchURLs []*url.URL) jsonnet.Importer {
...
 func (importer *universalImporter) Import(importedFrom, importedPath string) (jsonnet.Contents, string, error) {
 	log.Debugf("Importing %q from %q", importedPath, importedFrom)
 
+	binary := false
+	if strings.HasPrefix(importedPath, "binary://") {
+		if !importer.alpha {
+			return jsonnet.Contents{}, "", fmt.Errorf(`"binary://" url prefix requires the --alpha flag`)
+		}
+		binary = true
+		importedPath = strings.TrimPrefix(importedPath, "binary://")
```
anguslees, line 88: .. Hrm.  I think this means a remote binary import looks like `import "binary://https://github.com/path"` ... ?

I think this probably needs to change to "https+binary://" or similar (at the scheme level) if we want this to graduate from alpha.

mkmik: Yeah i wanted this hack out so my sre team at work could exercise the helmChart feature asap

mkmik: I still have hopes the importbin statement gets released...

anguslees: > I think this probably needs to change to "https+binary://" or similar (at the scheme level) if we want this to graduate from alpha.

Just in case anyone acts on this comment, `https+binary` scheme is also a bad idea, since the scheme is 'sticky' for relative imports, and will be incorrect.  I think the "correct hack" is a URL fragment/anchor, which only impacts the way the client _interprets_ the content, and not the way the server or relative URLs are constructed.

ie: _If_ we choose to keep this hack, we should make it: `import "https://github.com/path#binary"`

mkmik: Interesting!

Yeah we definitely need to change this before graduating this from alpha.

I still hope importbin goes through before that

PR #143, opened 2022-10-28
chore/importbin

primeroz:
Replace import binary:// with importbin in our tests


mkmik (approved):

utils/importer.go
```diff
@@ -98,6 +98,10 @@ func (importer *universalImporter) Import(importedFrom, importedPath string) (js
 	var tried []string
 	for _, u := range candidateURLs {
 		foundAt := u.String()
+		// Avoid collision bug when importing same chart in the same jsonnet file using `import binary://` and `importbin`
+		if binary {
+			foundAt = "binary-" + u.String()
```
mkmik, line 103: what happens if the imported file name starts with `binary-` ?

primeroz: it breaks, let me see to do it more cleanly by using the original binary:// prefix

primeroz: i changed it to a suffix like `##binaryImport` to the filename. i think the risk of someone files being called `CHARTNAME.tgz##binaryImport` are close to 0 :)

mkmik (approved):
it's better; although one could have escaped `#` in the url; but it's very very unlikely to clash and we're going to deprecate the `binary:/` hack soon

PR #150, opened 2022-11-22
Use gc tags present in input

muxmuse:
kubecfg supports garbage collection with a tag given by `--gc-tag`. This PR adds another flag `--gc-tags-from-input` which instructs kubecfg to garbage collect for all tags present on objects in the input.

This allows to deploy multiple applications at once where each application has its own gc tag.


mkmik:
So as long as at least one object remains, we can now purge old objects without having to fiddle too much with the command line. I think this is a reasonable trade off.

I'll review the code details ASAP

pkg/kubecfg/update.go
```diff
@@ -512,7 +525,23 @@ func eligibleForGc(obj metav1.Object, gcTag string) bool {
 		strategy = GcStrategyAuto
 	}
 
-	// [gctag-migration]: Check *label* == tag instead in phase2
-	return a[AnnotationGcTag] == gcTag &&
-		strategy == GcStrategyAuto
+	if strategy == GcStrategyAuto && a[AnnotationGcTag] != "" {
```
mkmik, line 528: I think this is a good opportunity to make this new feature use the `kubecfg.ksonnet.io/garbage-collect-tag` *label* and not the annotation.

I've been a bit "slow" (euphemism) to implement the migration from annotation to labels, but eventually this migration will happen.

It would be nice if we don't have to worry about people who checked in jsonnet in the VCSs with  `annotations: {'kubecfg.ksonnet.io/garbage-collect-tag': 'foo'}` without the corresponding label.

mkmik (commented):
Sorry for the late review!

Thanks for the PR; only one minor change request if you have time (see inline comment)

pkg/kubecfg/update.go
muxmuse: Hello @mkmik, Thank you for reviewing! I've changed the code such that the label is preferred over the annotation.

Playing around, I observed problems with metric-server's podmetrics which get labels of the pods they belong to, but no owner reference.

mkmik: > Playing around, I observed problems with metric-server's podmetrics which get labels of the pods they belong to, but no owner reference.

I don't think I understood, could you please describe the problem a bit more in details?

(sorry I was a bit busy; I should be a bit more responsive now)

mkmik: wow. So basically we need labels in order to search but we need also use the annotation to further filter the resultset

mkmik:
I'm going to merge this and see about the label propagation issue independently

mkmik (approved):

PR #157, opened 2022-12-07
Add a basic Dockerfile

muxmuse:
This PR adds a Dockerfile for an alpine based image with the kubecfg executable.


mkmik:
Thanks.

I'd prefer to use https://github.com/ko-build/ko

The main selling points are: no dependency on docker required ro build and multi-arch container cross compilation for free

mkmik:
> Example: https://github.com/mkmik/generated-secrets/blob/e2454fa6a86e97be5f01e1073cad8254e3148b07/.github/workflows/image-gh.yml#L21

actually a better example is https://github.com/mkmik/pigox/blob/main/.github/workflows/go.yml

muxmuse:
Thank you! I'll check it out and adjust this PR.

muxmuse:
Dockerfile has been replaced by a github workflow step that runs `ko build` after "Upload kubecfg artifact". This is a small increment as the image is only tagged with the commit hash and not with a version. It might still be helpful as an official source of kubecfg images.

mkmik (approved):

mkmik:
will fix CI permissions

mkmik:
ok, the images are available:

```bash
docker pull ghcr.io/kubecfg/kubecfg/kubecfg:latest
```

I'll figure out release tagging at next release

PR #166, opened 2022-12-21
Fix eval TLA

mkmik:
The `eval` command needs to run some user provided jsonnet expressions _in the context of_ a given file.
This works only if the "context" file returns an object. This doesn't work if the top level expression of a file is a function.

This means eval doesn't work with Top Level Arguments; which are a common way to parameterize jsonnet files.

This adds support for top level arguments:

```console
$ cat dummy.jsonnet                                               
function(a) {
  b: a * 2,
} 
$ ./kubecfg --alpha eval dummy.jsonnet --tla-code a='2'  
b: 4
```

---

The implementation trick is to craft a new top level function with one argument for each argument mentioned by one of the TLA flags and then just invoke the top level function sourced from the "context" file and pass down the arguments.

### DIFF
--- a/pkg/kubecfg/eval.go
+++ b/pkg/kubecfg/eval.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/google/go-jsonnet"
 	"github.com/kubecfg/kubecfg/utils"
@@ -17,7 +18,7 @@ type EvalCmd struct {
 	Format   string
 }
 
-func (c EvalCmd) Run(ctx context.Context, vm *jsonnet.VM, path string) error {
+func (c EvalCmd) Run(ctx context.Context, vm *jsonnet.VM, path string, tla []string) error {
 	expr := c.Expr
 	if expr == "" {
 		expr = "$"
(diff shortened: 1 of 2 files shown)


pkg/kubecfg/eval.go
```diff
@@ -26,7 +27,21 @@ func (c EvalCmd) Run(ctx context.Context, vm *jsonnet.VM, path string) error {
...
-	eval := fmt.Sprintf("((import %q) { Z____expr__Z_:: %s}).Z____expr__Z_", pathURL, expr)
+	var eval string
+	if err != nil {
+		return err
+	}
+	if len(tla) > 0 {
+		formals := strings.Join(tla, ",")
+		pairs := make([]string, len(tla))
+		for i := range tla {
+			pairs[i] = fmt.Sprintf("%s=%s", tla[i], tla[i])
```
lukebond, line 38: i don't really follow exactly what's happening here but this looks suspicious to me:

`fmt.Sprintf("%s=%s", tla[i], tla[i])`

should it be the same value either side of the equals sign?

mkmik: added a comment in the code

mkmik: basically the trick is `function(foo,bar) (import %q)(foo=foo,bar=bar)`

lukebond (approved):

PR #178, opened 2023-01-16
Implement oci: url scheme

mkmik:
```console
$ cat /tmp/imp.jsonnet                    
import 'oci://gcr.io/mkm-cloud/hello:v2'
$ make && ./kubecfg show /tmp/imp.jsonnet 
---
apiVersion: v1
data:
  foo: bar baz
kind: ConfigMap
metadata:
  name: demo
```

It implements a simple in-memory caching to avoid having to repeatedly interact with the remote OCI registry in the even of multiple relative imports from the same OCI package.

Like with `http://`, `oci://` works only in actual imports from within the jsonnet and the CLI arguments don't support urls (yet)

In order to create a package you have to use the `oras` tool:

```console
$ cat /tmp/demo.json
{
    "entrypoint": "demo.jsonnet"

$ tree /tmp/demo
/tmp/demo
├── data.txt
└── demo.jsonnet
$ cat /tmp/demo/demo.jsonnet
{
  config: {
    apiVersion: 'v1',
    kind: 'ConfigMap',
    metadata: {
      name: 'demo',
    },
    data: {
      foo: importstr 'data.txt',
    },
  },
}
$ cat /tmp/demo/data.txt
bar baz
$ (cd /tmp/demo; oras push gcr.io/mkm-cloud/hello:v2 --config /tmp/demo.json:'application/vnd.kubecfg.bundle.config.v1+json' .:'application/vnd.kubecfg.bundle.tar+gzip')
```

TODO in another PR:
- make oci paths work from the CLI
- implement a `kubecfg --alpha pack` command, so we don't have to use oras CLI to make packages (and remember all the details like the mime types)

### DIFF
--- a/utils/oci.go
+++ b/utils/oci.go
@@ -1,0 +1,164 @@
+// Copyright 2023 The kubecfg authors
+//
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package utils
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/StalkR/httpcache"
+	"github.com/containerd/containerd/remotes"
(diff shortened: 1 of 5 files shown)


utils/oci.go
```diff
@@ -0,0 +1,164 @@
...
+	"time"
+
+	"github.com/StalkR/httpcache"
+	"github.com/containerd/containerd/remotes"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/pkg/auth/docker"
+)
+
+const (
+	OCIBundleBodyMediaType   = "application/vnd.kubecfg.bundle.tar+gzip"
```
jaymebrd, line 37: Will it always be gzipped? I saw your comment about `kubecfg --alpha pack` so I guess it won't matter.

jaymebrd (approved):

utils/oci.go
mkmik: yes, oras builds a targz and the current code parses the content of the blob using gzip.Reader and tar.Reader

PR #195, opened 2023-02-15
Refactor common eval flags

mkmik:
This is no-effect refactoring that will pave the way for adding more common flags to all the commands that evaluate some jsonnet  (update,show,eval,delete,validate)

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -31,11 +31,13 @@ const (
 )
 
 func init() {
-	RootCmd.AddCommand(evalCmd)
-	evalCmd.PersistentFlags().StringP(flagExpr, "e", "", "jsonnet expression to evaluate")
-	evalCmd.PersistentFlags().BoolP(flagShowKeys, "k", false, "instead of rendering an object, list it's keys")
-	evalCmd.PersistentFlags().StringP(flagFormat, "o", "yaml", "Output format.  Supported values are: json, yaml")
-	evalCmd.PersistentFlags().String(flagExec, "", "Inline code") // like `jsonnet -e`
+	cmd := evalCmd
+	RootCmd.AddCommand(cmd)
+	cmd.PersistentFlags().StringP(flagExpr, "e", "", "jsonnet expression to evaluate")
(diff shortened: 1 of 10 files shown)


cmd/eval.go
```diff
@@ -31,11 +31,13 @@ const (
...
-	evalCmd.PersistentFlags().BoolP(flagShowKeys, "k", false, "instead of rendering an object, list it's keys")
-	evalCmd.PersistentFlags().StringP(flagFormat, "o", "yaml", "Output format.  Supported values are: json, yaml")
-	evalCmd.PersistentFlags().String(flagExec, "", "Inline code") // like `jsonnet -e`
+	cmd := evalCmd
+	RootCmd.AddCommand(cmd)
+	cmd.PersistentFlags().StringP(flagExpr, "e", "", "jsonnet expression to evaluate")
+	cmd.PersistentFlags().BoolP(flagShowKeys, "k", false, "instead of rendering an object, list it's keys")
+	cmd.PersistentFlags().StringP(flagFormat, "o", "yaml", "Output format.  Supported values are: json, yaml")
+
+	addCommonEvalFlags(cmd.PersistentFlags(), withoutShortEvalFlag())
```
jdockerty, line 40: Maybe I'm missing some wider context here, but how come this is called with `withoutShortEvalFlag`, when I believe `cmd.PersistentFlags().StringP(flagExpr, "e", ...` allows it to be called as `-e` for the shorthand expression anyway?

jdockerty (approved):
Seems like a good refactor to me.

Single comment, although I don't believe it is a blocker at all :smile:

cmd/eval.go
mkmik: yeah, there are two flags "-e", one is the "--exec/-e" used in all commands and the other is the "--eval/-e" used by the experimental (alpha) command `eval`.

I think this is confusing and we should have `-e` mean only `--exec` and find another way to convey the expression for `kubecfg eval`. Luckily the eval command is alpha, so we can change that later. However I don't want to unnecessarily break backward compatibility for this alpha command quite yet in this PR.

PR #196, opened 2023-02-15
Add --overlay flag

mkmik:
Implements an alpha feature that allows to apply overlays more easily. Demo:

```console
$ export KUBECFG_ALPHA=1
$ ./kubecfg show ./testdata/configmap.jsonnet
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
$ ./kubecfg show ./testdata/configmap.jsonnet --overlay ./testdata/namespace-overlay.jsonnet
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
  namespace: myns
```

This is basically a CLI syntactical sugar for:

```console
$  ./kubecfg show -e "(import './testdata/configmap.jsonnet') + (import './testdata/namespace-overlay.jsonnet')"
```

(which in turn is syntactical sugar for `kubecfg show 'data:,%28import%20%27...'`)

It works for all commands that accept jsonnet arguments (show,update,delete,validate,...).

If there are multiple jsonnet files, the same overlay is applied to all of the files. If you want more control over which overlay is applied on which file, you probably want to create intermediary files that import the source and the overlay

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -16,11 +16,15 @@
 package cmd
 
 import (
+	"fmt"
+
 	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
 )
 
 const (
-	flagExec = "exec"
+	flagExec    = "exec"
+	flagOverlay = "overlay"
 )
 
 type commonFlagOpts struct {
@@ -49,6 +53,7 @@ func addCommonEvalFlags(flags *flag.FlagSet, opt ...commonEvalFlagOpt) {
 		shortEval = ""
 	}
 	flags.StringP(flagExec, shortEval, "", "Inline code") // like `jsonnet -e`
+	flags.String(flagOverlay, "", "Jsonnet file to compose to each of the input files")
 }
 
 func processCommonEvalFlags(flags *flag.FlagSet, args *[]string) error {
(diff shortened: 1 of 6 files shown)


cmd/show_test.go
```diff
@@ -226,5 +226,61 @@ func TestShowExec(t *testing.T) {
 	if got != want {
 		t.Fatalf("got: %q, want: %q", got, want)
 	}
+}
+
+func TestShowOverlay(t *testing.T) {
```
jdockerty, line 231: Could this be converted into a table-test to reduce the boilerplate code around the multiple instances of `got := ... want := ...`

Not a huge deal since it is still clear what is going on and it looks like there is some revisiting anyway with the `TODO(mkm) ...` comment :blush:

cmd/common.go
```diff
@@ -59,5 +64,22 @@ func processCommonEvalFlags(flags *flag.FlagSet, args *[]string) error {
...
 	}
+
+	overlay, err := flags.GetString(flagOverlay)
+	if err != nil {
+		return err
+	}
+	if overlay != "" {
+		alpha := viper.GetBool(flagAlpha)
+		if !alpha {
+			return fmt.Errorf("--overlay is an alpha feature please use --alpha")
```
jdockerty, line 75: Minor nitpick:

Since this is `Errorf`, but with no format directive could we use the flag variables in their relevant places?

Granted, the more I think about that the uglier it actually looks, just a thought either way :laughing:

jdockerty (approved):
LGTM

cmd/show_test.go
mkmik: I'm a huge fan of table driven tests but in this case I think it doesn't pay the cost of its boilerplate.

cmd/common.go
mkmik: yeah, it's a tradeoff; it's not about ugliness, it's more about code searchability vs ensured consistency. I'm accepting your suggestion since it's indeed possible that alpha flags like this get renamed before stabilizing.

lukebond (approved):
:tada:

PR #212, opened 2023-03-06
feat: allow literal duplicates

cannonpalms:
Duplicate detection currently completely disallows duplicates of any kind based on (GroupKind, name, namespace). It can sometimes be useful to allow _literal_ duplicates, that is, duplicates where the objects' contents match completely. For example, two CRDs with identical definitions are not harmful, and so long as the CRD chosen is stable, we're in good shape.


utils/duplicates.go
```diff
@@ -24,13 +26,27 @@ import (
...
 		}
-		seen[k] = struct{}{}
+		seen[k] = v
 	}
 	return nil
 }
+
+func hash(obj *unstructured.Unstructured) string {
+	h := sha1.New()
+	// ignore error based on the unvalidated assumption that we already have a valid, marshallable unstructured object
```
cannonpalms, line 47: Is this assumption valid? I'd prefer not to have to deal with the pointer dereferencing required in order to make this function return `(*string, error)`.

mkmik: Yes we have marshallable unstructured objects

utils/duplicates.go
```diff
@@ -24,13 +26,27 @@ import (
...
+	json, _ := obj.MarshalJSON()
```
mkmik, line 48: I'm totally fine with this, however you could rely on the fact that hash implements a writer and Unstructured implements  custom marshaller interface:

```go
json.NewEncoder(h).Encode(obj)
```

utils/duplicates.go
mkmik: It's ok to not return the error but at least we could panic if we do receive an error (for some reason we cannot foresee here)

cannonpalms: Sure.

mkmik (approved):

utils/duplicates.go
cannonpalms: PTAL

utils/duplicates.go
cannonpalms: PTAL

mkmik (approved):

mkmik:
It fails with provenance fields (because they are annotations inside the object):

```console
$ ./kubecfg show testdata/duplicates_literal.jsonnet
---
apiVersion: v1
data:
  some: value
kind: ConfigMap
metadata:
  name: foo
  namespace: myns
---
apiVersion: v1
data:
  some: value
kind: ConfigMap
metadata:
  name: foo
  namespace: myns
$ ./kubecfg show testdata/duplicates_literal.jsonnet --show-provenance
ERROR duplicate resource ConfigMap, "myns", "foo"
```

mkmik (changes requested):
Needs to work with `--show-provenance` too.

mkmik:
(the test fails because the CI for some reason pushes the docker image to the GHCR repository even if this is just a PR; and coming from a fork it fails)

mkmik (changes requested):
:-(

it doesn't work with `--export-dir`:

```console
$ ./kubecfg show testdata/duplicates_literal.jsonnet --show-provenance --export-dir /tmp/bah
ERROR open /tmp/bah/v1.ConfigMap-myns.foo.yaml: file exists
```

instead of just ignoring the duplicate, we should be deduplicating the list of objects and keep only the first one (and ignore duplicates provided that the content is the same, module the provenance annotations)

cannonpalms:
> instead of just ignoring the duplicate, we should be deduplicating the list of objects and keep only the first one (and ignore duplicates provided that the content is the same, module the provenance annotations)

Ahh, I thought we were doing that already. Happy to do it.

cannonpalms:
I refactored `CheckDuplicates` into a `RemoveDuplicates` variant that returns a deduplicated slice of unstructured objects.

mkmik (approved):
LGTM

mkmik:
there are some tests that fail:

```console
$ make test
go test  -tags netgo -installsuffix netgo -ldflags="-X main.version=dev-2023-03-07T19:53:02+0100-dcb42464 " -race ./...
?   	github.com/kubecfg/kubecfg	[no test files]
?   	github.com/kubecfg/kubecfg/lib	[no test files]
?   	github.com/kubecfg/kubecfg/pkg/kubecfg/vars	[no test files]
?   	github.com/kubecfg/kubecfg/pkg/oci	[no test files]
--- FAIL: TestShowOverlay (0.00s)
    show_test.go:60: Running args [--alpha show ../testdata/configmap.jsonnet ../testdata/secret.jsonnet]
    show_test.go:265: got: "---\napiVersion: v1\nkind: Secret\nmetadata:\n  name: test\n---\napiVersion: v1\nkind: ConfigMap\nmetadata:\n  name: test\n", want: "---\napiVersion: v1\nkind: ConfigMap\nmetadata:\n  name: test\n---\napiVersion: v1\nkind: Secret\nmetadata:\n  name: test\n"
FAIL
FAIL	github.com/kubecfg/kubecfg/cmd	0.093s
ok  	github.com/kubecfg/kubecfg/pkg/kubecfg	0.206s
ok  	github.com/kubecfg/kubecfg/utils	0.751s
FAIL
make: *** [gotest] Error 1
```

cannonpalms:
Hmm...they passed on my machine. I'll take a look--thanks!

cannonpalms:
Yeah, they still pass on my machine. They also pass [in CI](https://github.com/kubecfg/kubecfg/actions/runs/4357365590/jobs/7616951697#step:6:1). What revision are you using?

cannonpalms:
After clearing my local test cache with `go clean -testcache`, they indeed still pass. Non-deterministic test, wrong revision, or environment difference?

cannonpalms:
It looks like that test failure you shared is caused by the order of output objects differing from what is expected.

### Got
```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: test
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test

```

### Want
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
---
apiVersion: v1
kind: Secret
metadata:
  name: test

```

It appears that I may have introduced some non-determinism with this change. I'll investigate.

cannonpalms:
Ok, fixed, and that cleaned up the code a bit too. I wasn't a big fan of the `objectHashPair` type. Now it's unnecessary!

PTYAL (hopefully the last). Great catch, by the way. It might be worth adding a test for stable output order in a future PR.

mkmik (approved):

PR #218, opened 2023-03-19
Add support for #sha256:... fragment

mkmik:
```console
$ ./kubecfg show "file://$PWD/testdata/configmap.jsonnet#sha256:7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f22"     
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
$ ./kubecfg show "file://$PWD/testdata/configmap.jsonnet#sha256:7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f23"
ERROR error reading file:///Users/mkm/p/kubecfg/testdata/configmap.jsonnet#sha256:7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f23: content hash mismatch: got: "7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f22", want "7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f23"
```

Works for any url type.

You can use it to ensure that content you download is actually immutable:

```console
$ ./kubecfg show 'https://raw.githubusercontent.com/kubecfg/kubecfg/v0.29.1/testdata/configmap.jsonnet#sha256:'

ERROR error reading https://raw.githubusercontent.com/kubecfg/kubecfg/v0.29.1/testdata/configmap.jsonnet#sha256:: content hash mismatch: got: "7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f22", want ""
$ ./kubecfg show 'https://raw.githubusercontent.com/kubecfg/kubecfg/v0.29.1/testdata/configmap.jsonnet#sha256:7871c163d6071da0289f4f20276f9e2a728ad2f82bd7e68361f986b53a585f22'

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
```

This can be also useful if you want to use more readable symbolic tag names instead of github commit hashes (since tags are not necessarily immutable but often when used as release tags they are intended to be immutable)

TODO: tests

### DIFF
--- a/utils/importer.go
+++ b/utils/importer.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


anguslees:
Nice, I like it.

I think we should decide now that "unknown" fragment keys are ignored, so we have backwards compatibility of url fragments.

Down that line of thinking: should we use query string encoding (key=value&key=value)? Just since that is an adjacent standard for key/value pairs and URL escaping...

mkmik:
Yes using query parameter encoding for the fragment too is a good idea, I will change it.

What should we do when:

1. A fragment doesn't confirm to this syntax? Silently ignore because of backward compat or issue a warning?
2. A fragment is well formed but contains keys that don't mean anything for us, should we warn or silently ignore?

 In future there will be need for new hash types like sha512 or whatnot. Clients predating this PR will just ignore this fragment and thus silently ignore the hash check. Clients following this PR but predating the introduction of sha512 (or whatnot) would have the opportunity to warn. But, should they warn?

PR #224, opened 2023-03-28
rfc: Flags From Files

mkmik:
See all details in [RFC](../tree/flagspec/docs/rfcs/rfc02.md)


docs/rfcs/rfc02.md
```diff
@@ -0,0 +1,89 @@
...
+
+I've seen a lot of people doing (2), with a wrapper script that sets up all the import-path flags
+to their vendored dependencies, passing ext vars etc etc.
+
+Once the user does that, it effectively puts vital config information into a language that not particularly well suited
+for describing configs, namely, shell scripts or makefiles or whatnot.
+
+## Proposal
+
+Right before starting up, search for a `.kubecfgrc.jsonnet` in the current dir, and recursively in the parent directories.
```
mkmik, line 33: current dir is easy, but is this what we really want?

We seem to want to have a foolproof way to just evaluate a file anywhere and have it pick up the right "flags" file. That would require to locate the rc file relative to the jsonnet file about to be evaluated.

Unfortunately most kubecfg commands support multiple jsonnet files.
We could locate the kubecfgrc.jsonnet files once per input jsonnet file and apply the flags only for that file.

However, some of the flags affect the global command output and cannot be isolated on individual flags.

hmm....

mkmik: perhaps we could just search the `.kubecfgrc.jsonnet` file starting from each of the files to evaluate and error out if they differ?

mkmik: updated document with this proposal

PR #226, opened 2023-03-29
Support data: uris in imports too

mkmik:
`data:` URIs are supported only as hack when passed to the top level evaluation function.

A `data:` URI is a bit silly but can be nevertheless useful. I noticed this when I was trying to add a trivial impl of `--overlay-code`. The existing code allowed any URL to be added as overlay, it was natural to just convert a literal code snippet into a URL.

We can have an implementation  over --overlay-code and --overlay-code-file that doesn't require data URLs, but it highlights the weirdness of supporting data: urls on the top level paths, but not inside import (which are often used with generated code such as `import '<a>' + import '<b>')


utils/data_uri.go
```diff
@@ -0,0 +1,64 @@
...
+	content, err := url.PathUnescape(strings.TrimPrefix(pathURL, "data:,"))
+	if err != nil {
+		return "", "", err
+	}
+	cwd, err := os.Getwd()
+	if err != nil {
+		return "", "", err
+	}
+
+	foundAt, err := PathToURL(cwd)
```
anguslees, line 57: I know this is what the original code had, but I think this is incorrect, and could cause the cached data URL "contents" to collide with other data URLs.

I think `foundAt` should be the original data URL.

utils/data_uri.go
```diff
@@ -0,0 +1,64 @@
...
+	content, _, err := expandDataURL(req.URL.String())
+	if err != nil {
+		return nil, err
+	}
+
+	return simpleHTTPResponse(req, http.StatusOK, io.NopCloser(strings.NewReader(content))), nil
+}
+
+func expandDataURL(pathURL string) (string, string, error) {
+	content, err := url.PathUnescape(strings.TrimPrefix(pathURL, "data:,"))
```
anguslees, line 48: `TrimPrefix` here makes me nervous, because it will be silently ignored, resulting in odd content for things like `data:text/plain,`

I think this function should accept the `url.URL` (since the caller already has a parsed URL), and look at `u.Opaque` (so we've already split off the scheme).  Then a `[encoding, content] := strings.SplitN(url.PathUnescape(u.Opaque), ",", 2)` - I think it's ok to return an error if `encoding != ""`.

mkmik: oh yeah, I completely ignored the mimetype part. I updated also many other places where I just matched on `data:,` ignoring the possibility that the user would provide encodings to data uris.

utils/data_uri.go
mkmik: yeah it's complicated.

The `foundAt` is used as the base to resolve relative paths when the data uri content itself then contains imports.

The main use case for data urls is to allow users to pass inline code to flags like in:

```console
$ ./kubecfg show foo.yaml --overlay-code '{name:: (importstr "README.md")[2:9]}'
---
apiVersion: v1
kind: Foo
spec:
  name: kubecfg
```

Data uris here are just an implementation hack. We could have implemented this `--overlay-code` functionality differently. But it was quite natural: all commands like show, update, validate, diff etc know how to read a url, so all we had to do is to turn "some_url" "--overlay-code" "some_str" into a `data:,(import "some_url") + (some_str)` (proper escaping not shown here) and feed this new url into the same existing logic that would process the input url.

Since the main use case was to use data uris internally to handle CLI parameters, using `cwd` was natural, since the user's intent was captured: it was reasonable that relative paths present in the literal string passed to the command were resolved relative to the current directory.

However, here we're expanding the ability to use this data urls also inside other files, and so the relative path resolution should be relative to those files, like any other import.

So I guess I have to find a way to weave this context through

mkmik: just to be clear, if I set `foundAt = pathURL.String()` my example command above will return this error:

```
RUNTIME ERROR: Get "data:///README.md": invalid data url: missing ','
```

mkmik: but since we're using `foundAt` as a cache key in:

```
		if c, ok := importer.cache[foundAt]; ok {
			return c, foundAt, nil
		}
```

we clearly cannot return whatever was the `importedFrom` who imported the data url, as the new `foundAt`, we need to return something that uniquely captures the data url **and** also the location where relative paths should be resolved against.

mkmik: I see a couple of options:

1. using a custom media type parameter attribute name, like `data:;kubecfg_imported_from=someuriencodeduri,somedata`

2. using a custom uri scheme like `kubecfgdata:///?base=someuriencodeduri&data=someuriencodeddata`

PR #227, opened 2023-03-29
Systematize --overlay-* flags to follow pattern of --ext-var and --tla

mkmik:
The `--overlay` flag is an alpha feature that allows users to quickly apply an overlay on a jsonnet file, directly from the CLI.

During initial alpha testing of the feature, it became clear that we more often needed to apply an overlay where the jsonnet source of the overlay appears as literal code snippet in in the CLI flags (as opposed to applying an overlay from a file).

The rationale for that is that when you have overlays in files, it's more common to just import the file meant to be overlaid and then apply the overlay: `(import "base.jsonnet")  { foo: 1 }`.

That said, it's still useful to sometimes decouple the overlay from the files to overlay.

We have already an established flag pattern to distinguish "literal code" from "code from file", and we use it in the `--ext-*` and `--tla-*` flag families.

I propose to do the same with `--overlay-`, except there is no need to support `--overlay-str` since that would mean `base + "overlay"` which is not reasonable overlay to apply (technically it stringifies the base object and appends the string "overlay" right after the closing brace, which is not something anybody would actually want to do.

Thus I propose two flags;

*  `--overlay-code`: literal jsonnet code source
*  `--overlay-code-file`: pathname of a jsonnet file

And keep `--overlay` as an alias of `--overlay-code-file` but issue a deprecation warning.

The choice may appear a bit verbose, but it follows an established parameter and should be obvious to whoever has seen `--tla-code` etc

We should remove `--overlay` flag before promoting `--overlay-code-*` flags out of alpha.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -291,17 +291,40 @@ func readObjs(cmd *cobra.Command, paths []string, opts ...utils.ReadOption) ([]*
 		paths = append(paths, utils.ToDataURL(exec))
 	}
 
+	overlayCodeFile, err := flags.GetString(flagOverlayCodeFile)
+	if err != nil {
+		return nil, err
+	}
+
 	overlay, err := flags.GetString(flagOverlay)
 	if err != nil {
 		return nil, err
 	}
 	if overlay != "" {
+		// deprecated. pflag will print a warning
+		overlayCodeFile = overlay
+	}
+
+	if overlayCodeFile != "" {
 		alpha := viper.GetBool(flagAlpha)
 		if !alpha {
-			return nil, fmt.Errorf("--%s is an alpha feature please use --%s", flagOverlay, flagAlpha)
+			return nil, fmt.Errorf("--%s is an alpha feature please use --%s", flagOverlayCodeFile, flagAlpha)
 		}
-		opts = append(opts, utils.WithOverlayURL(overlay))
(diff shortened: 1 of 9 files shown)


cmd/root.go
```diff
@@ -291,17 +291,40 @@ func readObjs(cmd *cobra.Command, paths []string, opts ...utils.ReadOption) ([]*
...
+	if err != nil {
+		return nil, err
+	}
+
 	overlay, err := flags.GetString(flagOverlay)
 	if err != nil {
 		return nil, err
 	}
 	if overlay != "" {
+		log.Warnf("%q is deprecated, please use %q", flagOverlay, flagOverlayCodeFile)
```
cannonpalms, line 304: Is there perhaps some way to avoid duplicating this deprecation warning across every command?

cannonpalms: I am blind and should stop reviewing code tonight

cannonpalms (approved):
Nice work.

cmd/root.go
cannonpalms: I am blind and should stop reviewing code for the night. This only occurs once.

mkmik: you made me spend 2m more to find the native way in pflag to deprecated a flag

PR #235, opened 2023-03-31
Add caching + vendoring

mkmik:
See all details in [RFC](../tree/vendoring/docs/rfcs/rfc03.md)


mkmik:
CC: @gus, @primeroz, @jaymebrd, @lukebond

anguslees:
I really want this :)

This (as a way to handle vendor/ use case, as well as simple performance optimisation) has been on my wishlist ever since we thought of import-by-URL.

docs/rfcs/rfc03.md
```diff
@@ -0,0 +1,127 @@
...
+
+The simplest thing to do is to let the cache just grow and grow and let the user just clean it up when needed.
+
+The other strategies are:
+
+* size limit, delete least recently used
+* remove all unreferenced
+* ???
+
+#### Remove all unreferenced strategy
```
anguslees, line 93: Thanks for writing this up.  This seems way harder than a "remove older than X" subcommand, and as you point out it's impossible unless we can enumerate _all_ the jsonnet top-levels using the shared cache somehow.

anguslees: Hrm. Another version is change the above "write to first" to instead _always_ write to first.  That way we can 'rotate' the cache dirs by adding a new cache dir first, that reads from the existing cache dirs; run all kubecfg we care about with the same new cache dir; then eventually replace the old cache dir with the content of the new cache dir (or flip flags to achieve the same thing).

This suggests that `--cache-dir ./vendor` always updates `./vendor` with a _copy_ of files used by this repo, pulled from any existing homedir-wide cache if there's a cache hit.   I think that's what people expect from `./vendor`, although this also suggests each repo pulls down its own copy without ever storing a copy in home-dir cache - which could be undesirable if I had many such repos.  I guess a power user can create a hierarchy of `--cache-dir ./vendor --cache-dir ../other/vendor`, if they care about this.  Another option is to _always_ write to the homedir-wide cache (unless disabled) in addition to the first `--cache-dir`.

(Fwiw, my prior imaginary simplest-thing-that-could-work version of this whole feature just had a single `--cache-dir` that defaulted to `~/.cache/kubecfg`.  If you change the value to `./vendor` then you're also implicitly disabling your homedir cache.  No cache eviction, we just encourage `rm -r; rerun` if folks care.  Real simple, real dumb.  I still kind of like this option because it's so easy to describe.)

docs/rfcs/rfc03.md
```diff
@@ -0,0 +1,127 @@
...
+### CI
+
+When running in a CI, it's advisable to rely solely on the vendored files and break the build if the user
+accidentally forgot to commit new imports:
+
+```bash
+kubecfg --ci ....
+```
+
+The `--ci` flag is intentionally generic because it may govern other behaviour that is sensible in a CI environment.
```
anguslees, line 116: Fwiw, I always thought of `kubecfg` _as_ a CI-optimised tool 😅 - unlike `kubectl` which is optimised for interactive use.  So this `--ci` flag seems weird.  I think I would prefer a `--no-network` flag or something similarly specific.

(CI-optimised is the reason there was no homedir .kubecfgrc config file, for example - the assumption was that kubecfg is usually invoked via a CI/CD pipeline or other shell wrapper, and we _want_ a bunch of flags in that shell script, rather than singleton environment files.)

docs/rfcs/rfc03.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Filename escaping
```
anguslees, line 118: Urgh.  I like making the URLs "understandable", particularly if we're asking folks to rm -r files from the cache directly.

An alternative (that I don't like, but mentioning for completeness) is that we just base64 or sha hash the URL (including scheme, path separators, etc), and the original URLs become opaque filenames at the filesystem level.

docs/rfcs/rfc03.md
```diff
@@ -0,0 +1,127 @@
...
+
+Example:
+
+```bash
+kubecfg --cache-dir ./vendor --cache-dir ~/.cache/kubecfg show ....
+```
+
+Open questions:
+
+* should we always put the default cache at the end of the cache search path?
```
anguslees, line 64: I think we need a way to say "no cache".  I think we should have a `--no-user-cache` flag (defaults to _using_ a homedir cache), which internally just adjusts whether we put the default homedir cache at the end of the search path.

In particular this means we can do weird things like `--cache-dir ./vendor --no-user-cache`, and it still sort of makes sense.

docs/rfcs/rfc03.md
mkmik: I also think of kubecfg as a CI tool.

But I also use it locally while I develop a PR that will be eventually run also by the CI.

I found that in practice having to always use a wrapper she'll script that passes flags is minor UX nightmare and hence I created this other RFC: https://github.com/kubecfg/kubecfg/blob/flagspec/docs/rfcs/rfc02.md

That said, for the issue at hand, I think the need to choose what to do when you import new files that are not vendored.  I can think of two main options:

a. interactive UX is subservient to automated UX: if you add/modify an import expression in a source file and when you execute kubecfg that import is not vendored we would fail hard and force the user to explicitly run a command or set an env/flag and rerun kubecfg.

b. The interactive UX automatically updates the vendor dir as needed. When running on CI we have to specify some flag that inhibits kubecfg from filling the vendor dir in case the user forgot to check-in the change.

Option (a) may appear to be a bit annoying but it would offer the opportunity to force the user to run a "kubecfg vendor" command that would update the vendor dir and also prune it. This would provide some incentive for the developer to run the pruning logic; otherwise with (b) the developer may never run pruning and only add files as they get added as needed by running kubecfg.

docs/rfcs/rfc03.md
mkmik: I think I brainfarted above:

All we have to do is to decouple foundAt from the cacheKey.

docs/rfcs/rfc03.md
```diff
@@ -0,0 +1,127 @@
...
+
+The `--ci` flag is intentionally generic because it may govern other behaviour that is sensible in a CI environment.
+
+### Filename escaping
+
+Naming files in a way that is safe despite having users that use different OSs like linux, macos and windows is quite harder than what it looks like.
+
+Macos is case preserving but case insensitive (in all but the most weird setups some few people have). Some macos versions don't like the ":" char, some do.
+
+Windows has a different path separator, supports far less punctuation, requires file names to be valid utf-8, and last but not least: has a set of odd reserved names that just cannot appear as filenames, things like `CON`, `NUL` are not valid file names!
```
mkmik, line 124: https://www.fileside.app/blog/2023-03-17_windows-file-paths/

docs/rfcs/rfc03.md
```diff
@@ -0,0 +1,127 @@
...
+kubecfg --cache-dir ./vendor --cache-dir ~/.cache/kubecfg show ....
+```
+
+Open questions:
+
+* should we always put the default cache at the end of the cache search path?
+
+### Vendoring
+
+To enable vendoring the user needs to pass a custom location for the cache directory and choose a desirable cache pruning policy.
```
jenniferplusplus, line 68: The thing that I would really like is a package manager. But as I started writing that comment, it occurred to me that consumers could probably build their own out of these primitives. And the more jsonnet-ey solution might be to let them do that, with some guidance as to how.

The way I'm imaging that would work is to create something like a packages.jsonnet file, and that would import and expose all of the external dependencies. Then, as a matter of convention, the rest of the jsonnet project references the objects exposed by that file, and never directly imports their own external dependencies.

Like so
```jsonnet
//<project_root>/packages.jsonnet
{
    acmeAppSpec: import 'https://github.com/acme/exampleApp/<gitsha>'
}
```

```jsonnet
//<project_root>/env/production.jsonnet
{
    app: k8s.apps.v1.Deployment {
        spec+: (import '../packages.jsonnet').acmeAppSpec {
            // normal jsonnet things here
        }
    }
}
```

Then if something like this becomes the standard guidance, the rest of the tooling can be designed to support that. For instance, cache cleaning can remove unused references, and it's easy to know what's used or not based just on this one file. If a repo needs to have more than one set of external references for some reason, each set can be expressed with it's own packages files, and its own cache dir.

jenniferplusplus: I think this also supports more advanced usage, too. For example
```jsonnet
//<project_root>/packages.jsonnet
{
    acmeAppSpec: {
        v2Config:: { experimental: true }
        v1: import 'https://github.com/acme/exampleApp/<gitsha>/app.jsonnet',
        v2: (import 'https://github.com/acme/appTwoDotOh/<gitsha>/app.jsonnet') + self.v2Config,
    }
}
```

mkmik: Not quite sure if we need a "package manager". If we have a way to point to remote sources a tool like renovate bot can just update those references. If you don't want to have them scattered across your files, you can easily create a level of indirection:

```
$ cat packages.jsonnet
{
  foo: import 'https://github.com/acme/foo/v0.1.0/app.jsonnet',
  bar: import 'https://github.com/acme/bar/v0.3.0/app.jsonnet',  
}
$ cat something.jsonnet
{
  foo: (import 'packages.jsonnet').foo,
}
```

then managing the versions of your dependencies across your project is just a matter of updating `packages.jsonnet` and you have all foos point to the same foo, while at the same time allowing people to import two different versions of a dependency if they really need it (something is quite a pain nowadays with jsonnet bundler)

jenniferplusplus:
@mkmik I've been coming back to this RFC a lot recently, and I really would love to have some version of it. What needs to happen to get this implemented?

mkmik:
> @mkmik I've been coming back to this RFC a lot recently, and I really would love to have some version of it. What needs to happen to get this implemented?

I think the main open question for me was fleshing out the filename escaping strategy since it's something that it's very hard to change later.

A precise spec with some examples, possible as code 

I can take care of fitting the rest of the stuff that is more entangled with the internals (importer magic)

jenniferplusplus:
Is that something I can help with?

jenniferplusplus:
@mkmik I just opened a PR #436 to support the percent encoding you described in the RFC

PR #251, opened 2023-05-09
Add eval --trace (powered by ursonnet)

mkmik:
Add `--trace` flag to `kubecfg eval`.

It prints out the locations of all jsonnet expression that would affect a given field in the output YAML.

----

For example imagine you feel lost in our jsonnet forest but you know your way in the Kubernetes YAML files and all you want is to change a specific field in a specific resource. You know what you want to change but you have no idea where that resource "comes from", which are the various templates, overlays and functions that have contributed to rendering that resource.

So here's what you do: you first locate the resource in our generated YAML files, find root template and root field path in "provenance" annotations and run the new "--trace" option on kubecfg eval, possibly passing a more focused "path" (e.g. `.spec.selector`

It will output all the `<filename>:<linenumber>` locations that contain expressions that could possibly affect the value you're focusing at. This significantly narrows down the places you have to look at to find the "right" place to perform your change and it also provides you with a sensible "thread" though the template source code for exploring and understanding the templates.

The tool guarantees that any changes you make outside the fields listed in the trace cannot possibly affect the value you're tracing.

Demo:

```console
$ ./kubecfg eval --alpha  ../ursonnet/testdata/child.jsonnet  -e '$.deployment.spec.template.spec.containers_.foo.resources.limits.memory' --
trace
2Gi

/Users/mkm/p/ursonnet/testdata/config.libsonnet:4 
/Users/mkm/p/ursonnet/testdata/common.libsonnet:22 
/Users/mkm/p/ursonnet/testdata/common.libsonnet:23 
/Users/mkm/p/ursonnet/testdata/base.jsonnet:5 
$
```


cannonpalms:
Nice

cannonpalms (approved):

PR #284, opened 2023-07-13
feat: add `validateJSONSchema` native function

jdockerty:
Adds the `validateJSONSchema` functionality to `kubecfg` so that a schema can be provided against a particular object. If this validation fails, a stream of errors are produced to `stderr` - under the common `RUNTIME ERROR: ...`.

This is a little rough around the edges, so any feedback is appreciated :smile:

### DIFF
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -66,7 +66,7 @@
   // validateJSONSchema(obj, schema): Validates a given object against the provided
   // schema. Returns 'true' is the schema is valid. If this is not the case, an error stream
   // is omitted based on the given schema's rules.
-  validateJSONSchema(obj, schema):: std.native('validateJSONSchema')(obj, schema),
+  validateJSONSchema:: std.native('validateJSONSchema'),
 
   // isK8sObject(o): Return true iff o is a Kubernetes object.
   isK8sObject(o):: (
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -16,78 +16,83 @@
 {
   // parseJson(data): parses the `data` string as a json document, and
   // returns the resulting jsonnet object.
-  parseJson:: std.native("parseJson"),
+  parseJson:: std.native('parseJson'),
 
   // parseYaml(data): parse the `data` string as a YAML stream, and
   // returns an *array* of the resulting jsonnet objects.  A single
   // YAML document will still be returned as an array with one
   // element.
-  parseYaml:: std.native("parseYaml"),
+  parseYaml:: std.native('parseYaml'),
 
   // manifestJson(value, indent): convert the jsonnet object `value`
   // to a string encoded as "pretty" (multi-line) JSON, with each
   // nesting level indented by `indent` spaces.
   manifestJson(value, indent=4):: (
-    local f = std.native("manifestJson");
+    local f = std.native('manifestJson');
     f(value, indent)
   ),
 
   // manifestYaml(value): convert the jsonnet object `value` to a
   // string encoded as a single YAML document.
-  manifestYaml:: std.native("manifestYaml"),
+  manifestYaml:: std.native('manifestYaml'),
 
   // escapeStringRegex(s): Quote the regex metacharacters found in s.
   // The result is a regex that will match the original literal
   // characters.
-  escapeStringRegex:: std.native("escapeStringRegex"),
+  escapeStringRegex:: std.native('escapeStringRegex'),
 
   // resolveImage(image): convert the docker image string from
   // image:tag into a more specific image@digest, depending on kubecfg
   // command line flags.
-  resolveImage:: std.native("resolveImage"),
+  resolveImage:: std.native('resolveImage'),
 
   // regexMatch(regex, string): Returns true if regex is found in
   // string. Regex is as implemented in golang regexp package
   // (python-ish).
-  regexMatch:: std.native("regexMatch"),
+  regexMatch:: std.native('regexMatch'),
 
   // regexSubst(regex, src, repl): Return the result of replacing
   // regex in src with repl.  Replacement string may include $1, etc
   // to refer to submatches.  Regex is as implemented in golang regexp
   // package (python-ish).
-  regexSubst:: std.native("regexSubst"),
+  regexSubst:: std.native('regexSubst'),
 
   // parseHelmChart(chartData, releaseName, namespace, values): Expand
   // helm chart into jsonnet objects.  `chartData` should be valid
   // chart .tgz as an array of numbers (bytes).  `values` is a jsonnet
   // object that conforms to a schema defined by the chart.
-  parseHelmChart:: std.native("parseHelmChart"),
+  parseHelmChart:: std.native('parseHelmChart'),
+
+  // validateJSONSchema(obj, schema): Validates a given object against the provided
+  // schema. Returns 'true' is the schema is valid. If this is not the case, an error stream
+  // is omitted based on the given schema's rules.
+  validateJSONSchema(obj, schema):: std.native('validateJSONSchema')(obj, schema),
 
   // isK8sObject(o): Return true iff o is a Kubernetes object.
   isK8sObject(o):: (
     std.isObject(o) &&
-    std.objectHas(o, "apiVersion") &&
-    std.objectHas(o, "kind")
+    std.objectHas(o, 'apiVersion') &&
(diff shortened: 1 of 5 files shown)


utils/nativefuncs.go
```diff
@@ -242,6 +243,33 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
...
+
+			jSchema, err := json.Marshal(schema)
+			if err != nil {
+				return nil, fmt.Errorf("unable to json marshal schema: %w\n", err)
+			}
+
+			// No URL defaults to draft 7 of JSONSchema
+			sch, err := jsonschema.CompileString("", string(jSchema))
+			if err != nil {
+				return false, fmt.Errorf("unable to compile jsonschema: %w\n", err)
```
jdockerty, line 262: Considering this means that the schema itself is invalid, i.e. a misplaced `required: [<fields>]` key I think this could also be `return nil, ...`.

Perhaps it should be, in order to avoid false negatives :thinking: 

**Thoughts?**

utils/nativefuncs_test.go
```diff
@@ -149,6 +150,56 @@ func TestParseHelmChart(t *testing.T) {
...
+
+    local obj = {
+        language: 'go',
+    };
+
+    std.native('validateJSONSchema')(obj, schema)`)
+	if err != nil {
+		if !strings.Contains(err.Error(), "invalid against the schema") {
+			t.Errorf("expected invalid schema error. got: %s\n", err.Error())
+		}
```
jdockerty, line 199: I'm not a huge fan of this, but I think it is _okay_ in a test. Any suggestions on this for a small test are appreciated :smile:

jdockerty (commented):
Small self-review for context/thoughts

mkmik:
> My editor also automatically runs jsonnetfmt on save, hence the larger diff for lib/kubecfg.libsonnet

Let's do this first and then rebase this PR on top of that: https://github.com/kubecfg/kubecfg/pull/286

mkmik:
https://github.com/kubecfg/kubecfg/pull/287 should help you fix the CI error.

@jdockerty btw you have write rights to this repo so you could create a PR directly here instead of via a fork

jdockerty:
> so you could create a PR directly here instead of via a fork

Ahh :facepalm: Useful to know for next time :smile:

utils/nativefuncs.go
```diff
@@ -242,6 +243,33 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
 			return ret, nil
 		},
 	})
+
+	vm.NativeFunction(&jsonnet.NativeFunction{
+		Name:   "validateJSONSchema",
+		Params: []jsonnetAst.Identifier{"obj", "schema"},
+		Func: func(args []interface{}) (interface{}, error) {
+			obj := args[0].(map[string]interface{})
+			schema := args[1].(map[string]interface{})
```
mkmik, line 252: what's the purpose of these "casts" ?

both `json.Marshal` and `.Validate` below accept `interface{}`

```suggestion
			obj := args[0]
			schema := args[1]
```

jdockerty: Agreed, I was going for readability, but there's no need to have the trailing `map[string]interface{}` casting.

utils/nativefuncs.go
mkmik: yeah returning nil on error is ok.

The reason we return `true` on success (as opposed to null) is to make the call to validateJSONSchema "assert-friendly".

```jsonnet
{
  assert kubecfg.validateJSONSchema(self, import 'schema.json'),

  // ...
}

File: utils/nativefuncs_test.go (hunk 2 above)
Comment (mkmik) [reply utils/nativefuncs_test.go:199]: it's ok

File: utils/nativefuncs.go [hunk 4]
```diff
@@ -242,6 +243,33 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
...
+
+			// No URL defaults to draft 7 of JSONSchema
+			sch, err := jsonschema.CompileString("", string(jSchema))
+			if err != nil {
+				return false, fmt.Errorf("unable to compile jsonschema: %w\n", err)
+			}
+
+			err = sch.Validate(obj)
+			if err != nil {
+				return false, fmt.Errorf("object is invalid against the schema: %w\n", err)
```
mkmik, line 267:
```suggestion
				return false, fmt.Errorf("object is invalid against the schema: %w", err)
```

utils/nativefuncs.go
```diff
@@ -242,6 +243,33 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
...
+
+			jSchema, err := json.Marshal(schema)
+			if err != nil {
+				return nil, fmt.Errorf("unable to json marshal schema: %w\n", err)
+			}
+
+			// No URL defaults to draft 7 of JSONSchema
+			sch, err := jsonschema.CompileString("", string(jSchema))
+			if err != nil {
+				return false, fmt.Errorf("unable to compile jsonschema: %w\n", err)
```
mkmik, line 262: anyway, let's remove newline in the error message (any linter would have told you that, unfortunately we're not running a linter in the CI job)

```suggestion
				return false, fmt.Errorf("unable to compile jsonschema: %w", err)
```

utils/nativefuncs.go
```diff
@@ -242,6 +243,33 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
...
+	vm.NativeFunction(&jsonnet.NativeFunction{
+		Name:   "validateJSONSchema",
+		Params: []jsonnetAst.Identifier{"obj", "schema"},
+		Func: func(args []interface{}) (interface{}, error) {
+			obj := args[0]
+			schema := args[1]
+
+			jSchema, err := json.Marshal(schema)
+			if err != nil {
+				return nil, fmt.Errorf("unable to json marshal schema: %w\n", err)
```
mkmik, line 256:
```suggestion
				return nil, fmt.Errorf("unable to json marshal schema: %w", err)
```

utils/nativefuncs.go
jdockerty: Makes sense :+1: 

The `assert` explanation helps too, that formalises more what we discussed before. I'll swap this to `return nil, ...`

lib/kubecfg.libsonnet
```diff
@@ -63,6 +63,11 @@
   // object that conforms to a schema defined by the chart.
   parseHelmChart:: std.native('parseHelmChart'),
 
+  // validateJSONSchema(obj, schema): Validates a given object against the provided
+  // schema. Returns 'true' is the schema is valid. If this is not the case, an error stream
+  // is omitted based on the given schema's rules.
+  validateJSONSchema(obj, schema):: std.native('validateJSONSchema')(obj, schema),
```
mkmik, line 69: The function arity and the names of the formal parameters are defined in Go:

```
	vm.NativeFunction(&jsonnet.NativeFunction{
		Name:   "validateJSONSchema",
		Params: []jsonnetAst.Identifier{"obj", "schema"},
```

so we don't need a jsonnet function that calls a native function, we can just return the native function (like the other entries above)

```suggestion
  validateJSONSchema:: std.native('validateJSONSchema'),
```

The advantages are:

1. no need to duplicate the names of the formal parameters
2. avoid having a stack trace entry that points to kubecfg.libsonnet

i.e. error messages are cleaner:

```
ERROR error reading /tmp/validate.jsonnet: RUNTIME ERROR: object is invalid against the schema: jsonschema: '' does not validate with file:///Users/mkm/p/kubecfg#/additionalProperties: additionalProperties 'somethingelse' not allowed
        file:///tmp/validate.jsonnet:5:12-77    object <anonymous>
        Checking object assertions
        Field "spec"
        During manifestation
```

as opposed to

```
ERROR error reading /tmp/validate.jsonnet: RUNTIME ERROR: object is invalid against the schema: jsonschema: '' does not validate with file:///Users/mkm/p/kubecfg#/additionalProperties: additionalProperties 'somethingelse' not allowed
        internal:///kubecfg.libsonnet:69:37-82  function <anonymous>
        file:///tmp/validate.jsonnet:5:12-77    object <anonymous>
        Checking object assertions
        Field "spec"
        During manifestation
```

jdockerty: Makes sense!

mkmik (approved):
LGTM

PR #289, opened 2023-07-17
Initial support for DOCS publishing to github pages

primeroz:
This PR add initial support, and initial documentation, to be built using `mkdocs-material` and be pushed to `github pages` ( **We need to enable github pages in the repo settings ) 

You can see the latest build of this code in my fork https://primeroz.github.io/kubecfg/ 

* Add initial docs, not all pages have been documented yet and some have a TODO list but i focuse mainly on the topics i was interest in at first like
  * helm integration
  * metacontroller integration
  * traceback
* Added a full example for metacontroller in `examples/metacontroller` 
* 2 new `make targets` for serving docs locally while building and building docs ( used in ci ) are added, they are only tested on linux

This is really a work in progress and going forward i will 

* finish the current pages marked as TODO  ( if i can figure out how the OCI one works ) 
* ideally we should add new docs pages when new features are added


mkmik (approved):
Thanks!

PR #296, opened 2023-07-19
Add flag to pull var from Environment with fallback to default value

primeroz:
add a new `--ext-str-env FOO=_default_value_` to allow looking up a variable from the environment and have a default if the value is missing

**TODO**
- [ ] Update Docs
- [x] Test

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -62,6 +62,7 @@ const (
 	flagTLACodeFile = "tla-code-file"
 	flagResolver    = "resolve-images"
 	flagResolvFail  = "resolve-images-error"
+	flagExtVarEnv   = "ext-str-env"
 )
 
 var clientConfig clientcmd.ClientConfig
@@ -89,6 +90,7 @@ func init() {
 	RootCmd.MarkPersistentFlagFilename(flagTLACodeFile)
 	RootCmd.PersistentFlags().String(flagResolver, "noop", "Change implementation of resolveImage native function. One of: noop, registry")
 	RootCmd.PersistentFlags().String(flagResolvFail, "warn", "Action when resolveImage fails. One of ignore,warn,error")
+	RootCmd.PersistentFlags().StringArray(flagExtVarEnv, nil, "Values of external variables with string values to fetch from the process environment with a default value `ENV=default`")
 
 	// The "usual" clientcmd/kubectl flags
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -62,6 +62,7 @@ const (
 	flagTLACodeFile = "tla-code-file"
 	flagResolver    = "resolve-images"
 	flagResolvFail  = "resolve-images-error"
+	flagExtVarEnv   = "ext-str-env"
```
jdockerty, line 65: Very minor thing, but for organisation I suppose we should move this up, alongside the other `ext-str` flags

https://github.com/kubecfg/kubecfg/blob/d40858b0f72ba725c422d8cc82d6251f336d1538/cmd/root.go#L55-L58

cmd/root.go
```diff
@@ -271,7 +274,15 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 						return nil, fmt.Errorf("Missing environment variable: %s", kv[0])
 					}
 				case 2:
-					spec.setter(kv[0], kv[1])
+					if spec.flagName == flagExtVarEnv {
+						if v, present := os.LookupEnv(kv[0]); present {
```
jdockerty, line 278:
```suggestion
						if v, ok := os.LookupEnv(kv[0]); ok {
```

Not a huge deal, the `ok` for the `bool` in the map is the accepted Go convention for this.

jdockerty:
This LGTM, I'll give it some time before the full :heavy_check_mark: so you can read the other comments and give mkm sometime to review as well.

---

As an extra, I wonder whether it'll be worthwhile to expand onto the docs for the use-case of this (I see there is a TODO for that as well :+1:), as I might be missing the full intention.

I used a very small dummy example to test it, e.g. :point_down: 

```jsonnet
// test.jsonnet
{
  apiVersion: 'v1',
  kind: 'Pod',
  metadata+: {
    name: 'test',
    namespace: 'test',
    labels: {
      env: std.extVar('TEST_ENV'),
    },
  },
  spec: {},
}
```

In order to have that set, I would then do something like `TEST_ENV="prod" kubecfg show test.jsonnet --ext-str-env TEST_ENV=local`

This then renders the `env: prod` because I had it set, but if I didn't have it set then it be `local`. Maybe my test is bad here, but it seems odd to specify the `TEST_ENV` twice, i.e. the specified default could just be the value anyway - as I say though, perhaps I'm not fully understanding the intended use-case too. Perhaps if `kubecfg` is being ran across multiple environments there is always the intended default provided and it acts as a safeguard too?

mkmik:
also I don't like the fact we're breaking the symmetry. all other `*-str`, `*-code`, `*-str-file`, `*-code-file` interpret `key=value` as `key` is the variable name, and `value` is something sources the value (the `str|code` modifier specifies how the value is interpreted, whether the `-file` modifier interprets where the value comes from.

So for example `--ext-str-env varname=envvar` would be an option if we wanted to map an env var into a ext var of a different name. And that would mean we have to find a way to encode a default argument; it's common in shells to use `:-`:

```--ext-str-env foo=FOO:-default```

That if the ability to support a default is important. Otherwise I'd just supporting what jsonnet already supports when the value is missing ` -V / --ext-str <var>[=<val>] ` would be fine

primeroz:
> That if the ability to support a default is important. 

Yeah this was my goal, sorry to not make it clear. 

Essentially it was to simplify my life because once a `std.extVar('foo')` is defined in `jsonnet code` than the `--ext-str foo` must always be set or the jsonnet rendering will fail 

> --ext-str-env foo=FOO:-default

Maybe this is too confusing to be worthy 

I can handle this in the container definition so i can use the already provided `env lookup` functionality and will close this :+1:

PR #302, opened 2023-07-26
feat: kubecfg pack metadata. kubecfg version + custom metadata

mkmik:
NOTE: `kubecfg pack` is still behind `--alpha` feature flag. This is an experiment.

When creating an OCI package with `kubecfg pack`, we currently record only the filename of the entrypoint in the metadata.

It would be useful to contain other metadata but we don't know yet which metadata will be useful.

This PR adds a `metadata` field in the bundle config which will contain a JSON object which should have top level keys scoped by metadata type, possibly tool specific.

For example:

```json
{
  "entrypoint": "shell.jsonnet",
  "metadata": {
    "kubit.kubecfg.dev/v1alpha1": {
      "minKubecfgVersion": "v0.30.0"
    },
    "pack.kubecfg.dev/v1alpha1": {
      "version": "v0.31.4"
    }
  }
}
```

The custom metadata is obtained by evaluating an `$._kubecfg_pack_metadata` expression on the root jsonnet and gracefully handling when that field doesn't exist.

### DIFF
--- a/cmd/pack_test.go
+++ b/cmd/pack_test.go
@@ -178,7 +178,7 @@ func TestPush(t *testing.T) {
 	// these digests have been observed by running the test failing and obverving the logs
 	const (
 		blobDigest = "sha256:90847d8bd5ca990ebf777bf131130cdcf695ace2f8bb59dee0571129af855f00"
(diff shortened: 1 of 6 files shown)


jdockerty (approved):
LGTM

As you say, this is gated behind the `--alpha` flag anyway and is experimental. We can run with this as-is and alter as needed :rocket:

PR #303, opened 2023-07-26
feat: Allow users to specify where to apply the overlay

mkmik:
There are a few ways to pass parameters in kubecfg: TLA, ExtVars, Overlays.

The TLA and ExtVars mechanisms lend themselves to naturally ask you a "variable name" and a "value", separated by `=`.

AS for the overlay, though, it does make sense to apply the overlay on top of the whole object, hence we started by just asking a code literal or a filename.

However in practice it happens that it's very common to apply the overlay *inside* a specific child field in the top level object.

This PR adds support for specifying an optional "field path" in `--overlay-code` and `--overlay-code-file`:

Demo:

```console
$ cat /tmp/foo.jsonnet 
{
  a: {
    b: {
      apiVersion: 'v1',
      kind: 'ConfigMap',
      data: {},
    },
  },
}
$ go build && ./kubecfg --alpha show /tmp/foo.jsonnet  --overlay-code a.b='{data+:{foo:"bar"}}'
---
apiVersion: v1
data:
  foo: bar
kind: ConfigMap
$ cat /tmp/overlay.jsonnet 
{data+:{foo:"bar"}}
$ go build && ./kubecfg --alpha show /tmp/foo.jsonnet  --overlay-code-file a.b=/tmp/overlay.jsonnet
---
apiVersion: v1
data:
  foo: bar
kind: ConfigMap
```

---

TODO(mkm): Tests!


pkg/kubecfg/vm.go
```diff
@@ -231,22 +232,45 @@ func dirURL(path string) *url.URL {
 	return &url.URL{Scheme: "file", Path: path}
 }
 
+func buildOverlayObject(format, flag string) string {
+	re := regexp.MustCompile(`^([^$=]*)=(.*)$`)
+	parts := re.FindStringSubmatch(flag)
```
jdockerty, line 237: `FindStringSubmatch` can also return `nil`, I'm not sure whether we actually need to handle that case though considering the passed `flag` automatically fails if you give it nothing.

Raising awareness to it nonetheless :smile:

jdockerty (approved):

pkg/kubecfg/vm.go
mkmik: `len(slice)` when the slice is nil yields 0

PR #304, opened 2023-07-27
test: Add test for --overlay-* path=value

mkmik:
Tests for: https://github.com/kubecfg/kubecfg/pull/303

### DIFF
--- a/pkg/kubecfg/vm_test.go
+++ b/pkg/kubecfg/vm_test.go
@@ -12,7 +12,7 @@ import (
 func tempFile(t *testing.T, body string) (filename string) {
 	t.Helper()
 
-	tmpfile, err := os.CreateTemp("", "kubecfg-test-*.jsonnet")
+	tmpfile, err := os.CreateTemp(t.TempDir(), "kubecfg-test-*.jsonnet")
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -23,9 +23,6 @@ func tempFile(t *testing.T, body string) (filename string) {
 	if err := tmpfile.Close(); err != nil {
 		t.Fatal(err)
 	}
-	t.Cleanup(func() {
-		os.Remove(tmpfile.Name())
-	})
 	return tmpfile.Name()
 }
 
--- a/pkg/kubecfg/vm_test.go
+++ b/pkg/kubecfg/vm_test.go
@@ -1,0 +1,89 @@
+package kubecfg
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"testing"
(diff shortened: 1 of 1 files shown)


pkg/kubecfg/vm_test.go
```diff
@@ -0,0 +1,89 @@
...
+	obj := objs[0]
+	t.Logf("obj: %q", obj)
+
+	if got, want := obj.GetObjectKind().GroupVersionKind().Kind, "ConfigMap"; got != want {
+		t.Fatalf("got: %q, want: %q", got, want)
+	}
+	return obj.GetAnnotations()
+}
+
+func TestReadObjec(t *testing.T) {
```
jdockerty, line 56:
```suggestion
func TestReadObject(t *testing.T) {
```

pkg/kubecfg/vm_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if _, err := io.WriteString(tmpfile, body); err != nil {
+		tmpfile.Close()
+		t.Fatal(err)
+	}
+	if err := tmpfile.Close(); err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() {
+		os.Remove(tmpfile.Name())
+	})
```
jdockerty, line 28:
```suggestion
	tmpfile, err := os.CreateTemp(t.TempDir(), "kubecfg-test-*.jsonnet")
	if err != nil {
		t.Fatal(err)
	}
	if _, err := io.WriteString(tmpfile, body); err != nil {
		tmpfile.Close()
		t.Fatal(err)
	}
	if err := tmpfile.Close(); err != nil {
		t.Fatal(err)
	}
```

Thoughts on doing something like this?

I.e. we pass in [`t.TempDir()`](https://pkg.go.dev/testing#T.TempDir) which is from the `testing` stdlib that will automatically provide a directory that will be cleaned up after the test anyway - allowing us to slightly simplify the setup and remove our own call to `t.Cleanup`.


> TempDir returns a temporary directory for the test to use. The directory is automatically removed by Cleanup when the test and all its subtests complete.

jdockerty (changes requested):
Mainly for the typo :smile: 

Otherwise LGTM :rocket:

pkg/kubecfg/vm_test.go
mkmik: Ah yeah I forgot about that helper

mkmik:
suggestions accepted; PTAL

jdockerty (approved):

PR #308, opened 2023-07-28
feat: add renderOverlay

jdockerty:
This adds the `renderOverlay` function in order to wrap the `toOverlay` functionality which returns the resulting object of having the overlay applied.


testdata/kubecfg_test.jsonnet
```diff
@@ -78,6 +78,25 @@ local result =
...
+  };
+  local expectedOverlayObj = {
+    a: {
+      b: {
+        c: 2,
+        d: 10,
+      },
+    },
+  };
+  std.assertEqual(kubecfg.renderOverlay(testRenderOverlay, import 'overlay.json'), expectedOverlayObj) &&
```
jdockerty, line 97: This doesn't actually work at the moment, I'm trying to figure out why :thinking: 

```
kubecfg show kubecfg_test.jsonnet
ERROR error reading kubecfg_test.jsonnet: RUNTIME ERROR: Field does not exist: renderOverlay
```

mkmik (commented):
I don't think a function like that is useful as stdlib

testdata/kubecfg_test.jsonnet
jdockerty: Nevermind, I'm dumb :facepalm: 

I'm using `kubecfg` compiled myself which doesn't have the updated `lib/kubecfg.libsonnet`

jdockerty:
@mkmik Do you think the `toOverlay` should be modified to return the resulting object instead?

I.e. `toOverlay(obj, overlay)`

PR #310, opened 2023-08-02
feat(parseHelmChart): support passing capabilities

jaredallard:
Updates `parseHelmChart` to support passing `capabilities` to be passed as `chartutil.Capabilities` to the helm chart renderer. Right now, only `KubeVersion` is able to be passed. This is important for rendering helm charts that conditionally render `apiVersion`s, among other fields, based on the current Kubernetes Cluster. The [default is `1.20`](https://github.com/helm/helm/blob/f31d4fb3aacabf6102b3ec9214b3433a3dbf1812/pkg/chartutil/capabilities.go#L31) which doesn't work for all use cases.

I wasn't sure if this is the best approach, nor is this currently an `optional` field (I will figure that out here soon). I wanted to send this out as proposal PR as well as to get feedback. We need this where I work currently, so I'm very motivated to get this working!

Thank you ✨

### DIFF
--- a/utils/nativefuncs.go
+++ b/utils/nativefuncs.go
@@ -165,12 +165,27 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
 
 	vm.NativeFunction(&jsonnet.NativeFunction{
 		Name:   "parseHelmChart",
-		Params: []jsonnetAst.Identifier{"releaseName", "namespace", "chartData", "values"},
+		Params: []jsonnetAst.Identifier{"releaseName", "namespace", "chartData", "values", "capabilities"},
 		Func: func(args []interface{}) (interface{}, error) {
 			chartData := args[0].([]interface{})
 			releaseName := args[1].(string)
 			namespace := args[2].(string)
 			vals := args[3].(map[string]interface{})
+			mapCaps := args[4].(map[string]interface{})
+
+			// If capabilities are set, process them. Use the default
+			// capabilities as the base. If a user provided capabilities, only
+			// copy over the KubeVersion as the other fields are not easily
+			// configurable through the jsonnet interface or, in the case of the
+			// helm version, don't make sense.
+			var capabilities = chartutil.DefaultCapabilities.Copy()
+			if capKubeVersion, ok := mapCaps["KubeVersion"].(map[string]interface{}); ok {
+				capabilities.KubeVersion = chartutil.KubeVersion{
+					Version: capKubeVersion["Version"].(string),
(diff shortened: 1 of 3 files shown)


jaredallard:
Hmmm, it looks like this can't be made an optional parameter ([ref](https://github.com/google/go-jsonnet/blob/v0.20.0/thunks.go#L265)). It seems that native functions aren't able to set that (I may be wrong, so please correct me if my reading of the go-jsonnet source is incorrect.

jaredallard:
> the jsonnet wrapper around the native calls can do default parameters

Good call! I'll look into that soon!

utils/nativefuncs.go
```diff
@@ -165,13 +165,31 @@ func RegisterNativeFuncs(vm *jsonnet.VM, resolver Resolver) {
...
+			// helm version, don't make sense.
+			var capabilities = chartutil.DefaultCapabilities.Copy()
+			if len(args) >= 4 {
+				mapCaps := args[4].(map[string]interface{})
+
+				if capKubeVersion, ok := mapCaps["KubeVersion"].(map[string]interface{}); ok {
+					capabilities.KubeVersion = chartutil.KubeVersion{
+						Version: capKubeVersion["Version"].(string),
+						Major:   fmt.Sprintf("%s", capKubeVersion["Major"]),
+						Minor:   fmt.Sprintf("%s", capKubeVersion["Minor"]),
```
jaredallard, line 188: I know `fmt.Sprintf` is generally a smell, but this seemed like the easiest approach to converting into a string when the value could be `int` or `float`. Happy to change this, but I do feel the performance penalty doesn't warrant doing so.

mkmik: The problem is that `%s` will complain if the value of the interface is not a string (e.g. it will write something like `%!s(int=42)`).

You can either use `%v` or just use `fmt.Sprint(capKubeVersion["Major"])` which will stringify any value

jaredallard: I totally meant to do `%v`... sigh. Thank you!

mkmik (approved):
Thanks!

jaredallard:
I'll fix the tests here soon!

jaredallard:
@mkmik I rebased to remove the extra commits (since it looks like y'all do merge commits). The tests should be passing now (they are locally!)

Let me know if there's anything else I should do! Thanks!!

mkmik:
sorry, feel free to ping me if I forget about things.

jaredallard:
@mkmik Rebased again to remove the merge commits (it _seemed_ like that might've been the cause of the previous status failure)

PR #311, opened 2023-08-04
feat: initial support for `kustomize+https` import scheme

jdockerty:
Related to https://github.com/kubecfg/kubecfg/issues/219, I thought this was quite interesting so I spent a little bit of time with it.

This is a _very rough_ implementation of adding `kustomize+https://<url>` support to `kubecfg`. It should also be likely marked as requiring `--alpha`, which I can look into later on. No doubt there is a lot to do to get this working properly.

I.e. at a high-level this simply runs `kustomize build <url>` and relies entirely on their libraries. Perhaps in the future we want to also support `kustomize` locally with `kustomize://path/to/kustomization` or similar. For now, this works over a network.

**Note: as this is an initial start, it is missing the vendored/tagged portion, i.e. this will always get `master|main` at the moment. However, It should serve as a starting point to develop further.**

---

## Example

```jsonnet
// testdata/kustomize/test.jsonnet
{
  metacontrollerUpstream: import 'kustomize+https://github.com/metacontroller/metacontroller/manifests/production',
}
```

```bash
go run main.go show testdata/kustomize/test.jsonnet | kubectl apply -f -
namespace/metacontroller created
serviceaccount/metacontroller created
clusterrole.rbac.authorization.k8s.io/metacontroller created
clusterrolebinding.rbac.authorization.k8s.io/metacontroller created
clusterrole.rbac.authorization.k8s.io/aggregate-metacontroller-view created
clusterrole.rbac.authorization.k8s.io/aggregate-metacontroller-edit created
customresourcedefinition.apiextensions.k8s.io/compositecontrollers.metacontroller.k8s.io created
customresourcedefinition.apiextensions.k8s.io/controllerrevisions.metacontroller.k8s.io created
customresourcedefinition.apiextensions.k8s.io/decoratorcontrollers.metacontroller.k8s.io created
statefulset.apps/metacontroller created

```

### DIFF
--- a/utils/kustomize.go
+++ b/utils/kustomize.go
@@ -1,0 +1,37 @@
+package utils
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+
+	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+// kustomizeImporter satifies the http.RoundTripper interface
+type kustomizeImporter struct{}
+
+// RoundTrip performs a HTTP transaction for an `import kustomize+https://<url>` statement
+// by calling a simple Kustomize run against it, returning the rendered manifests.
+func (k *kustomizeImporter) RoundTrip(req *http.Request) (*http.Response, error) {
+
+	// We know the scheme is kustomize+https, so simply grab the URL
+    // for kustomize to use
+	url := strings.Split(req.URL.String(), "+")[1]
+
+	kustomizer := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	fs := filesys.MakeFsOnDisk()
+	m, err := kustomizer.Run(fs, url)
+	if err != nil {
+		return nil, err
+	}
+
+	yamlDocs, err := m.AsYaml()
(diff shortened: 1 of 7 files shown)


utils/kustomize.go
```diff
@@ -0,0 +1,37 @@
...
+	url := strings.Split(req.URL.String(), "+")[1]
+
+	kustomizer := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	fs := filesys.MakeFsOnDisk()
+	m, err := kustomizer.Run(fs, url)
+	if err != nil {
+		return nil, err
+	}
+
+	yamlDocs, err := m.AsYaml()
```
mkmik, line 31: if instead of yaml we output JSON, we can `import` and avoid the `std.parseYaml(importstr ...`

jdockerty: Makes sense, I put this here since the `ResMap` interface has the `AsYaml` func, but this simply returns `[]byte`, so I can have this as JSON instead.

Good idea :100:

mkmik (approved):

jdockerty:
cc: @primeroz As I know you were interested in this too :blush:

PR #321, opened 2023-08-14
fix: gate kustomize support behind alpha feature flag

jdockerty:
This was not previously behind the `--alpha` feature flag, when it should have been.

Now, we have an error when we do not use the `--alpha` feature flag with this, which is expected.

```
go run main.go show testdata/kustomize/test.jsonnet
ERROR error reading testdata/kustomize/test.jsonnet: RUNTIME ERROR: Get "kustomize+https://github.com/metacontroller/metacontroller/manifests/production": kustomize+https:// prefix is an alpha feature, please use the --alpha flag
        file:///home/influx/work/kubecfg/testdata/kustomize/test.jsonnet:2:27-115     object <anonymous>
        Field "metacontrollerUpstream"
        During manifestation
exit status 1

go run main.go show testdata/kustomize/test.jsonnet --alpha | head
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    app.kubernetes.io/name: metacontroller
  name: metacontroller
---
apiVersion: v1
...etc
```

### DIFF
--- a/utils/importer.go
+++ b/utils/importer.go
@@ -61,7 +61,7 @@ func MakeUniversalImporter(searchURLs []*url.URL, alpha bool) jsonnet.Importer {
 	t.RegisterProtocol("file", http.NewFileTransport(http.Dir("/")))
 	t.RegisterProtocol("internal", http.NewFileTransport(newInternalFS()))
 	t.RegisterProtocol("oci", newOCIImporter())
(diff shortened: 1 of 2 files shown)


mkmik:
yeah, alternatively we could just not register the protocol handler if alpha features are not enabled

PR #342, opened 2023-10-04
feat: add `getPath`/`objectHasPath`/`objectHasPathAll` functions for access to nested objects

jdockerty:
This adds the following functions:
- `getPath(obj, path, default=null, inc_hidden=true)`
- `objectHasPath(obj, path, inc_hidden=false)`
- `objectHasPathAll(obj, path)`

into the `kubecfg.libsonnet` file.

This is shorthand for needing to do something which looks like this for nested objects:

```
if std.objectHas(myObj, "foo") && std.objectHas(myObj.foo, "bar") && ...
```

This was adapted from an [old jsonnet mailing list](https://groups.google.com/g/jsonnet/c/1nEJOYmS78I/m/uQ-g-9A3BwAJ) and morphed to be more inline with the standard library implementations for the respective `jsonnet` functions.

### DIFF
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -158,9 +158,7 @@
 
   local objectHasDeep(o, f) = (
     local objectHasDeep_(o, ks) =
-      if !(std.type(o) == 'object' || std.type(o) == 'array') then
-        false
-      else if !std.objectHasAll(o, ks[0]) then
+      if !std.objectHasAll(o, ks[0]) then
         false
       else if std.length(ks) == 1 then
         true
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -158,7 +158,9 @@
 
   local objectHasDeep(o, f) = (
     local objectHasDeep_(o, ks) =
-      if !std.objectHasAll(o, ks[0]) then
+      if !(std.type(o) == 'object' || std.type(o) == 'array') then
+        false
+      else if !std.objectHasAll(o, ks[0]) then
         false
       else if std.length(ks) == 1 then
         true
--- a/lib/kubecfg.libsonnet
+++ b/lib/kubecfg.libsonnet
@@ -156,9 +156,24 @@
     objectGetDeep_(o, std.split(f, '.'))
   ),
 
+  local objectHasDeep(o, f) = (
+    local objectHasDeep_(o, ks) =
+      if !std.objectHasAll(o, ks[0]) then
+        false
+      else if std.length(ks) == 1 then
+        true
+      else
+        objectHasDeep_(o[ks[0]], ks[1:]);
+
+    objectHasDeep_(o, std.split(f, '.'))
+  ),
+
   // Similar to std.get, but with a nested jsonpath.
   // If a value is not available, then a default is given
   getAtPath(obj, path, default):: objectGetDeep(obj, path, default),
+
+  // Similar to std.objectHasAll, but with a nested jsonpath.
+  hasAtPath(obj, path):: objectHasDeep(obj, path),
 
   layouts:: {
     // gvkName(accum, o): Helper for 'fold'.  This accumulates a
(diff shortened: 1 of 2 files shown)


testdata/kubecfg_test.jsonnet
```diff
@@ -194,6 +194,18 @@ local result =
...
   std.assertEqual(two, { 'example.com/v1alpha1.Test': { _: { a: obj('a'), b: obj('b') } } }) &&
 
+  local nested_obj = {
+      foo:: {
+          bar:: {
+              baz:: 'nested!'
+          }
+      }
+  };
+  std.assertEqual(kubecfg.getAtPath(nested_obj, 'foo.bar.baz', 'default!'), 'nested!') &&
```
mkmik, line 204: how does it fail when we try something like:

```
kubecfg.hasAtPath({foo:10), 'foo.bar.baz'), true)
```

jdockerty: Ah yeah, I see what you mean. It assumes that the value is nested as an object at all times

```jsonnet
  std.assertEqual(kubecfg.hasAtPath({foo: 10}, 'foo.bar.baz'), false) && ...
```

Leads to
```
go run main.go show testdata/kubecfg_test.jsonnet --alpha
ERROR error reading testdata/kubecfg_test.jsonnet: RUNTIME ERROR: Unexpected type number, expected object
```

jdockerty: @mkmik Thoughts on [this addition](https://github.com/kubecfg/kubecfg/pull/342/commits/cc5e89a46afa11a35cfcdd1631593305bd33e138) to this?

I've added it as a work in progress commit so it can be reverted if so. It makes sense to me, but I'm also not sure whether the behaviour would be what we would desire.

I.e. `{foo: 10}` and trying to find `foo.bar.baz` _should_ be `false` because that path doesn't exist, compared to erroring out. Simply adding the `std.type` check for the edge case of passing in a small object like :point_up: 

Perhaps there's an edge case I'm missing, but I think either way works fine, i.e. get the `default` or error on the "semi-correct" path where the field is a `number`.

mkmik: to be fair, I don't know what it should be, I was just asking what would happen.

I think it's it's ok if it throws an error.

The default value is meant to to be useful when the data is missing, not when the data has the wrong shape.

jdockerty: > I think it's it's ok if it throws an error.

Agreed, I was thinking that using the wrong shape can also be interpreted as "missing", but I think erroring is maybe better in this situation too since it can be an issue on the caller's side.

Either works for sure, I can change it back. Your call on whichever you would prefer in the project :smile:

mkmik: yeah I think erroring is better

jdockerty: Reverted to the previous state :+1:

lib/kubecfg.libsonnet
```diff
@@ -144,6 +144,37 @@
       error ('o must be an object or array of k8s objects, found ' + std.type(o))
   ),
 
+  local objectGetDeep(o, f, default) = (
```
mkmik, line 147: I gave it some thought and I think for consistency with the jsonnet stdlib we should do this:

- `kubecfg.getPath(o, p, default=null, inc_hidden=true)` would behave like `std.get` but with a path
- `kubecfg.objectHasPath(o, p, inc_hidden=true)` like `std.objectHas` but allowing the user to ignore hidden fields like in `std.get` `inc_hidden`
- `kubecfg.objectHasPathAll(o, p)` as a simple wrapper for `kubecfg.objectHasPath(o,p,inc_hidden=false)`

jdockerty: Whilst I was having lunch I thought of the first one as well :laughing: 

I'll tweak this a little bit and get it to fit more inline with the `std` implementations :+1:

mkmik (approved):
LGTM!

PR #357, opened 2023-10-26
rfc: local git urls

mkmik:


mkmik:
CC: @anguslees, @primeroz

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+```
+
+The "production" template thus imports both the "app" template and the "v1" inventory JSON file. The "production" template overlays the "app" template with the "v1" inventory config, effectively causing every deployment that imports the staging template into a deployment that runs "v1" of the software.
+
+
+## Problem
+
+The main problem with this setup is that while each "phase" can have some specific overlays, and we make use of that feature to use a different version of the main application binaries in staging and in production, we have only one version of the main templates.
+
+> Any change to the templates applies immediately on every target cluster.
```
cannonpalms, line 221:
```suggestion
> [!IMPORTANT]  
> Any change to the app templates applies immediately on every target cluster.
```

I think your intent was to call attention to this as a sort of "problem statement." This is Github-specific markdown, but would help with the callout. There are also "NOTE" and "WARNING" options.

image


I've also clarified the templates to which you're referring.

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+
+   In the same vein as when you have your application A that depends on your library L at version V1, and another application B that depends on your library L at version V2, the templates for cluster A depend on the templates from library L at version V1 and cluster B depends on the templates from library L at version V2.
+
+   Most programming languages have package managers that express dependencies that include name _and_ version. If we do the same with the dependency between the cluster templates and the _application templates_ we'd solve the [problem](#Problem).
+
+3. The reason we don't normally think in terms of (1) and (2) is that the cluster-specific Jsonnet entry points are small and the bulk of the templates live in the _library_. Because of this it goes against the grain to fully split the _applications_ (cluster entrypoints like our `deploy/prod/aws/us-east-1.jsonnet`) from the _library_ (the `app/app.jsonnet`). It would be too many repositories to handle. It's too much work, and we're drawn to think of it in another way
```
cannonpalms, line 257:
```suggestion
3. The reason we don't normally think in terms of (1) and (2) is that the cluster-specific Jsonnet entry points are small and the bulk of the templates live in the _library_. Because of this it goes against the grain to fully split the _applications_ (cluster entrypoints like our `deploy/prod/aws/us-east-1.jsonnet`) from the _library_ (the `app/app.jsonnet`) in separate repositories. It would be too many repositories to handle. It's too much work, and we're drawn to think of it in another way.
```

"_split how_" clarity for this sentence

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+
+4. The last insight is that we don't need to split in separate repos. All we need is a way to refer to old _revisions_ within the same repository.
```
cannonpalms, line 259: This is a meta-insight that refers to insight (3) rather than an insight borne of the "Problem" above. Perhaps it belongs outside the list with context from the list baked-in?
```suggestion

We don't need to separate cluster-specific entrypoints and the library of app templates into repositories. All we need is a way to refer to old _revisions_ within the same repository.
```

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+
+Luckily Jsonnet an extensible offers import hooks functionality that kubecfg already taps extensively.
+
+## Proposal
+
+We propose to add a `srcfs` URL scheme and an import hook implementation that will allow to import file from the current project's version control system as it was a specific revision.
```
cannonpalms, line 265:
```suggestion
We propose to add a `srcfs` URL scheme and an import hook implementation that will allow to import files from the current project's version control system as it was a specific revision.
```

cannonpalms, line 265: It wasn't immediately clear to me what `srcfs` would do; I had to read your description here. 

I think you mean `src` as in "source control." If you intend to leave the implementation open to version control systems other than git, then perhaps this could be `vcsfs`?

This could be a problem that others won't share with me, so please speak out if I'm alone here.

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+
+### Lifecycle
+
+Let's see how this feature is meant to be used in practice to continuously deploy the latest tested version of an app to production.
+
+Let's assume we have a given version of our application+config template bundle being deployed in staging and that we have some automated Post Deployment Tests telling us whether everything is healthy.
+
+All we need to do is to make an automated PR that "updates" the `history-channel/prod` file to point to whatever is the current "healthy" version in staging.
+
+> I spent quite some time fighting with mermaid to draw this and I ultimately gave up and drew it by hand. So even if you're not a visual person, I'm sure you'll admire the primary colors of the following diagrams.
```
cannonpalms, line 296: Excellent handwriting :wink:

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+
+```console
+$ git archive --format=tar --output=/dev/null $REVISION
+```
+
+### Vendoring
+
+[RF03 Caching and Vendoring](https://github.com/kubecfg/kubecfg/blob/vendoring/docs/rfcs/rfc03.md) describes a general mechanism for vendoring external references.
+
+Should `srcfs` be considered an internal or an external reference?
```
cannonpalms, line 447:
```suggestion
#### Should `srcfs` be considered an internal or an external reference?
```

I thought you were presenting an open question at first.

docs/rfcs/rfc04.md
```diff
@@ -0,0 +1,551 @@
...
+    stag-aws-us-west-1(us-west-1)-->stag-aws-common
+    stag-aws-common(common)-->aws & stag
+
+    stag-gcp-us-east-1(us-east-1)-->stag-gcp-common
+    stag-gcp-us-west-1(us-west-1)-->stag-gcp-common
+    stag-gcp-common(common)-->gcp & stag
+
+    prod-->app
+    stag-->app
+```
```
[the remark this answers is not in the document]
mkmik: I choose the graph orientation to mimic the file tree orientation; the arrows mean "imports" and they go to the left.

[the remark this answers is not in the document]
mkmik: ah perhaps you're saying that the mermaid source could be written using the other orientation

docs/rfcs/rfc04.md
mkmik: yeah; the reason the name is srcfs is a reference/allusion to google's internal tool; in pretty much the same way as kubecfg is also an allusion.

I think I'll move to something more explicit like `git+file:` or something `local-git:`

PR #370, opened 2023-11-17
fix: typo iff to if

Peefy:
fix: typo iff to if


mkmik (approved):
Yeah "iff" is often a way to say "if and only if" but in this case it's not really necessary to put emphasis on that.

PR #436, opened 2024-05-09
feat(vendoring): add percent-encoding util to ensure safe cross-platform filenames

jenniferplusplus:
Note this targets the vendoring branch, to support #235

### DIFF
--- a/utils/filename.go
+++ b/utils/filename.go
@@ -46,7 +46,7 @@ func escapeAll(s string) string {
 	var buf strings.Builder
 
 	for i := 0; i < len(s); i++ {
-		buf.WriteString(fmt.Sprintf("%%%02x", s[i]))
+		fmt.Fprintf(&buf, "%%%02x", s[i])
 	}
 
 	return buf.String()
--- a/utils/filename.go
+++ b/utils/filename.go
@@ -1,0 +1,59 @@
+package utils
+
+import (
+	"regexp"
+	"strings"
+)
+
+const lowerHex = "0123456789abcdef"
+
+var allowedChars = regexp.MustCompile(`[^0-9a-z!@#$&()\[\]{},.;~\-_=+]`)
+
+// These are reserved words and symbols which cannot be a file name, with or without a file extension
+// Mainly special Windows devices
+var reservedWords = regexp.MustCompile(`^aux|^com[0-9¹²³]|^con|^lpt[0-9¹²³]|^nul|^prn`)
+
+// Symbols that cannot (or should not) make up the entirety of a filename
+var illegalSolo = regexp.MustCompile(`^\.+$`)
+
+// Symbols that cannot (or should not) begin a file name
+var illegalPrefix = regexp.MustCompile(`^-`)
+
+// Symbols that cannot (or should not) end a file name
+var illegalSuffix = regexp.MustCompile(`\.$`)
+
+// Encode percent encodes a filename string, to minimize cross-platform compatibility issues related to case handling,
+// allowed characters, and reserved words
+func Encode(str string) (string, error) {
+	filename := allowedChars.ReplaceAllStringFunc(str, escape)
+	filename = illegalSolo.ReplaceAllStringFunc(filename, escape)
+	filename = illegalPrefix.ReplaceAllStringFunc(filename, escape)
+	filename = illegalSuffix.ReplaceAllStringFunc(filename, escape)
+
+	parts := strings.SplitN(filename, ".", 2)
+	parts[0] = reservedWords.ReplaceAllStringFunc(parts[0], escape)
+
+	return strings.Join(parts, "."), nil
+}
+
+// escape converts every byte in string s to a percent encoded string
+// similar to net/url methods, but simplified to remove any decision logic, and using lowercase letters
+func escape(s string) string {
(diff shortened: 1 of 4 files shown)


utils/filename_test.go
```diff
@@ -0,0 +1,139 @@
...
+		expected: "this%20has%20spaces",
+	},
+	{
+		scenario: "capitals",
+		given:    "This Has Caps And Spaces",
+		expected: "%54his%20%48as%20%43aps%20%41nd%20%53paces",
+	},
+	{
+		scenario: "emojis",
+		given:    "happy😀file",
```
mkmik, line 37: emojis in filenames work just fine on windows, macos and linux, no need to escape them

jenniferplusplus: I was using emojis as an easy stand in for the overall extended unicode character set. My concern was mainly about case sensitivity in russian or greek character sets, for example.

This was a demonstration that multi-byte and multi-code point graphemes will encode and decode correctly. I'm not sure how I would do that otherwise, unless you want to bring in dedicated unicode libraries.

utils/filename.go
```diff
@@ -0,0 +1,59 @@
...
+var allowedChars = regexp.MustCompile(`[^0-9a-z!@#$&()\[\]{},.;~\-_=+]`)
+
+// These are reserved words and symbols which cannot be a file name, with or without a file extension
+// Mainly special Windows devices
+var reservedWords = regexp.MustCompile(`^aux|^com[0-9¹²³]|^con|^lpt[0-9¹²³]|^nul|^prn`)
+
+// Symbols that cannot (or should not) make up the entirety of a filename
+var illegalSolo = regexp.MustCompile(`^\.+$`)
+
+// Symbols that cannot (or should not) begin a file name
```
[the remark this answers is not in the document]
jenniferplusplus: Starting with a file name with a dash is actually valid in all OSes, as far as I could tell. But, it seems to be very common that shells would handle that poorly and interpret the file name as a flag instead, which raises the possibility of shell injection. It seemed safer this way.

utils/filename_test.go
mkmik: I'm wondering if unicode filenames are going to be a problem or not at all in the first place.

I'm wondering if we're being excessively defensive here. Is there an example of files on repos on github that cannot be correctly `git clone`d on windows because of unicode characters that are file on linux or mac but not on windows?

Ideally we should mangle names the least amount of possible, in order to keep the vendored directory to feel as much as possible "native" and "reviewable" and reserve the escaping only for edge cases where we know people will be very confused by the behavior of the system due to illegal filenames.

I'm also trying to figure out if there is a better way to handle uppercase chars, which are going to the single most common class of escaped chars under the current scheme.

utils/filename.go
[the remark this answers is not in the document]
mkmik: the goal of the name mangling here is not to protect the user from bad file names but to provide a consistent experience to people who vendor files across OSs. Files in the vendor or cache directories should be rarely accessed by end users and more often they will just be noticed because they will appear in code reviews

utils/filename_test.go
mkmik: for example, what if instead of escaping uppercases, we implement some sort of conflict resolution pass, where we sanitize the entire list of files and we only escape things when there are actual conflicts? (not sure how to do that)

jenniferplusplus: I have an update that mangles only uppercase letters, and reserved words and symbols. So lowercase greek, cyrillic, and other character sets are left alone. In general, so are symbols that are not letter-like.

Let me know if you see any problems

utils/filename.go
```diff
@@ -0,0 +1,60 @@
...
+	return strings.Join(parts, "."), nil
+}
+
+// escapeAll converts every byte in string s to a percent encoded string
+// similar to net/url methods, but simplified to remove any decision logic, and using lowercase letters
+func escapeAll(s string) string {
+	var buf strings.Builder
+
+	for i := 0; i < len(s); i++ {
+		buf.WriteString(fmt.Sprintf("%%%02x", s[i]))
```
mkmik, line 49: Ah yeah you're right we need to iterate on bytes not runes.

That said, you can use Fprintf which is going to write more directly to the string builder

```suggestion
		fmt.Fprintf(&buf, "%%%02x", s[i])
```

jenniferplusplus: Fprintf returns an error, which is hard to deal with here. Unless you want me to just ignore it? The error seems to be about failing to write to the buffer, which shouldn't be an issue when the buffer is a strings.Builder.

mkmik: the error is caused by the underlying error returned by the io.Writer; in our case that's the string builder so it can be safely ignored

mkmik: (btw buf.WriteString also returns an error; it does so in order to implement  the https://pkg.go.dev/io#StringWriter interface)

jenniferplusplus: Huh, I wonder why Goland complained about the unhandled error from one, but not the other. 🤷🏻‍♀️

Anyway, I updated to use Fprintf.

jenniferplusplus:
@mkmik Just checking in to see if there's any other changes you want from this?

PR #442, opened 2024-08-20
fix: set resolver in jsonnet vm

tompizmor:
Since kubecfg version 0.29.1 (more precisely since [this commit](https://github.com/kubecfg/kubecfg/commit/6541e5b07ac767fbbaab59c10a13c687c5b77de6) IIFNM) the resolve-images feature stopped working since the resolver was not set.

I noticed the `flagResolver` and `flagResolvFail` flags were simply ignored and I also saw the existing `WithResolver` function was not used at all.

I tested this MR in my environment and it works fine. Please let me know I case I have not considered something important.


pkg/kubecfg/vm.go
```diff
@@ -83,6 +84,18 @@ const (
...
+	ResolverType_name = map[ResolverType]string{
+		0: "noop",
+		1: "registry",
+	}
+	ResolverType_value = map[string]ResolverType{
+		"noop":     0,
+		"registry": 1,
+	}
+)
+
```
jotadrilo, line 98: Let's use specific enum values for easy the reading:

```suggestion
// Enum value maps for ResolverType.
var (
	ResolverType_name = map[ResolverType]string{
		NoopResolver: "noop",
		RegistryResolver: "registry",
	}
	ResolverType_value = map[string]ResolverType{
		"noop":     NoopResolver,
		"registry": RegistryResolver,
	}
)

```

pkg/kubecfg/vm.go
```diff
@@ -91,6 +104,20 @@ const (
...
+		1: "warn",
+		2: "error",
+	}
+	ResolverFailureAction_value = map[string]ResolverFailureAction{
+		"ignore": 0,
+		"warn":   1,
+		"error":  2,
+	}
+)
+
```
jotadrilo, line 120: Let's use specific enum values here too

cmd/root.go
```diff
@@ -237,6 +237,10 @@ func JsonnetVM(cmd *cobra.Command) (*jsonnet.VM, error) {
 		}
 	}
 
+	resolverType := kubecfg.ResolverType_value[viper.GetString(flagResolver)]
+	resolverFailureAction := kubecfg.ResolverFailureAction_value[viper.GetString(flagResolvFail)]
```
jotadrilo, line 241: It is "weird" to access this kind of value mapping from a different package. I'd create a `Parse` method to parse the flag string and make the internal value mappings "private" (unexported).

#### Extra points

Consider having a method to list the enum's allowed string values and one to get the string of the enum type.

Then update the cmd flag definitions to use the `NoopResolver` string representation as the default flag value and the enum's allowed string values as part of the flag's description.

tompizmor: Done!

tompizmor:
@mkmik in the name of Bitnami, would you be so kind to take a look at this? 😄

jotadrilo (approved):
LGTM!

mkmik (approved):

tompizmor:
@mkmik could you release a new version to include this fix?

PR #448, opened 2024-11-01
feat: --annotation cli arg for pack command

### LINKED ISSUE #447: proposal: add annotations args to kubecfg pack command
i would like to [annotate](https://github.com/opencontainers/image-spec/blob/main/annotations.md) the OCI bundles i produce from kubecfg pack (to store provenance information, e.g. source git commit and such). i experimented with using `crane mutate` to set them on the bundle by tag (which creates a new image obviously, but moves the tag) but it clears the config that kubecfg pack puts on the bundle (or more accurately doesn't copy it over to the new bundle). this means you can't later render it with `kubecfg show` because the entrypoint is in the config. i believe crane does this because oras uses config for non-standard things that container type tools like crane don't expect.

as i see it the best solution is to support the addition of annotations in kubecfg pack itself. should simply be a matter of exposing the CLI arg and then putting them into the manifest (not the config).

happy to contribute it myself but thought i'd write up the background and rationale here.

lukebond:
closes #447 

e.g.:

```console
$ kubecfg --alpha pack \
  --annotation my-domain.com/my-repo-commit-sha=252e791916a082b075f6d84daf46ec8545482b7e648f8208f81f3f4e4c728a \
  --annotation my-domain.com/pull-request=https://github.com/my-org/my-repo/pull/123
```

### DIFF
--- a/cmd/pack.go
+++ b/cmd/pack.go
@@ -27,6 +27,7 @@ const (
 	flagOutput           = "output"
 	flagInsecureRegistry = "insecure-registry"
 	flagDocsTarFile      = "docs-tar-file"
+	flagAnnotations      = "annotation"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


mkmik (approved):
LGTM

PR #452, opened 2025-01-13
Add OCI resolution support

jotadrilo:
I found out that the digest resolution feature does not work for OCI manifests since the underlying utilities (github.com/genuinetools/reg) do not support OCI.

Specifically, this lib performs some remote requests to the container registry with the header `Accept: application/vnd.docker.distribution.manifest.v2+json` only.

https://github.com/genuinetools/reg/blob/master/registry/digest.go#L28
https://github.com/distribution/distribution/blob/main/manifest/schema2/manifest.go#L17

This lib would work with OCI manifests if the request includes the `application/vnd.oci.image.manifest.v1+json` media type too, but it is not the case.

This PR replaces the genuinetools utilities in favor of the Google's utilities for containers, which is more extended.

I have also simplified the local `ImageName` struct to remove unused information.

### Current status

```shell
❯ /usr/local/bin/kubecfg show --resolve-images=registry -V images_tag=a62debace8881fc4123cb8c4429c849f067a9c52-rlocal-jdrios my-service.jsonnet | grep -oE '@sha256:.*'
WARNING unable to get digest from the registry: invalid checksum digest format
```

### Fix

```shell
❯ ~/go/bin/kubecfg show --resolve-images=registry -V images_tag=a62debace8881fc4123cb8c4429c849f067a9c52-rlocal-jdrios my-service.jsonnet | grep -oE '@sha256:.*'
@sha256:06a310247695796a5d2e0cf5c14745301544a3a7b22d72b4a9a2bd7754caeb31
```


mkmik:
thanks.

There seems to be a regression in the output. 

```
error reading testdata/kubecfg_test.jsonnet: RUNTIME ERROR: Assertion failed. index.docker.io/library/busybox:latest != docker.io/library/busybox:latest
```

it's possibly caused by the different library you're using?

It's important to maintain backward compat; can you see if you can special case the docker.io output? (only if it's caused by this change and it's not because of an upstream change)

jotadrilo:
> thanks.
> 
> There seems to be a regression in the output.
> 
> ```
> error reading testdata/kubecfg_test.jsonnet: RUNTIME ERROR: Assertion failed. index.docker.io/library/busybox:latest != docker.io/library/busybox:latest
> ```
> 
> it's possibly caused by the different library you're using?
> 
> It's important to maintain backward compat; can you see if you can special case the docker.io output? (only if it's caused by this change and it's not because of an upstream change)

I actually fixed that just after I pushed the branch. See https://github.com/kubecfg/kubecfg/pull/452/commits/e5354edd7eb489dd83e1ba999ad1aee8a0112cd1

mkmik (approved):

jotadrilo:
@mkmik would you mind cutting a release to include this chagne? We are interested on moving to OCI manifests and keep the digest resolution on.

jotadrilo:
@mkmik friendly ping!

mkmik:
done

PR #453, opened 2025-02-01
Add `--qps-limit` flag for `rest.Config.QPS`

mmonaco:
## Add `--qps-limit` flag for `rest.Config.QPS`

Background: The k8s `client-go` package sets a default client-side QPS of 5.

When managing many resources (dozens) this rate limiting is very noticable.

I think it's reasonable for `kubecfg` to disable the rate-limiting by
default:

 - AFAICT a lot of the work is done sequentially anyway. Disabling this
   makes `kubecfg` snappier with many resources but cannot result in
   massive floods of parallel requests.

 - If `kubecfg` did do more in parallel, it would also make sense to
   disable the client side rate limiting and then set custom parallelism
   / work pool size / batch size.

However, to be conservative leave the default behavior.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -62,6 +62,7 @@ const (
 	flagTLACodeFile = "tla-code-file"
 	flagResolver    = "resolve-images"
 	flagResolvFail  = "resolve-images-error"
+	flagQPSLimit    = "qps-limit"
 )
 
 var clientConfig clientcmd.ClientConfig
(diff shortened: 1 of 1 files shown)


mmonaco:
> Thanks!
> 
> I don't know why the rate limiting is in place nor what would happen in practice if we remove it.
> 
> I think we should make it configurable. Probably default to enabled, so that people can opt-out if it's causing problems.

Done

mmonaco:
> > Thanks!
> > I don't know why the rate limiting is in place nor what would happen in practice if we remove it.
> > I think we should make it configurable. Probably default to enabled, so that people can opt-out if it's causing problems.
> 
> Done

Oh, now I see "default enabled" but TBH I'm not sure which direction you'd prefer. The current state in this PR is now that the library default of 5 QPS is left in place but we can disable/tune with the `kubecfg --qps-limit` flag.

mmonaco:
Friendly ping :)

mkmik (approved):

PR #454, opened 2025-03-07
ci: support linux/arm64 builds

charlesthomas:
i'm using `kubecfg` in a multi-platform image, built on self-hosted GitHub Actions runners. for some reason, one of the dependencies of `kubecfg` causes the `go install` to take a **very** long time (the whole image build takes 30 min; most of it is the `go install`).

if `kubecfg` published a linux arm build alongside the linux x64 build, then i could install both via `curl` instead of `go install`

refs:
  - [docs on linux arm runners](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#choosing-github-hosted-runners)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,7 +14,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest]
     runs-on: ${{ matrix.os }}
 
     env:
@@ -107,7 +107,7 @@ jobs:
   release:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest]
     runs-on: ${{ matrix.os }}
     needs: create_release
     if: startsWith(github.ref, 'refs/tags/v')


mkmik (approved):

mkmik:
sounds good. We have a cross-build of the docker image (with `ko`), so we could also do a cross build of the binary, but this is easier

charlesthomas:
> sounds good. We have a cross-build of the docker image (with `ko`), so we could also do a cross build of the binary, but this is easier

i noticed the image build, but i couldn't figure out if / where the image gets published. the docs have a "to do" item for the image. it looked to me like it was going to ghcr, but https://ghcr.io/kubecfg/kubecfg gave me a 404

mkmik:
For "reasons" the name of the image is: `ghcr.io/kubecfg/kubecfg/kubecfg`

mkmik:
```console
$ docker run --rm -ti ghcr.io/kubecfg/kubecfg/kubecfg:v0.35.2 version
kubecfg version: v0.35.2
jsonnet version: v0.20.0
client-go version: v0.0.0-master+$Format:%H$
```

mkmik:
anyway, fair point. I didn't know it was so hard to find the docker image.

I added a mention to the docker image location in the release notes. hopefully that will make it easy to find!

I released v0.36.0 which now includes an ARM binary you can download.

It would be nice if we had an official "setup-kubecfg" github action that people could use to quickly download the kubecfg binary from the release page, like many tools have. Would you like to help with that @charlesthomas  ?

PR #498, opened 2026-09-14
Remove dependency on oras.land/oras-go (v1).

dsymonds:
The v1 branch is no longer maintained, and has unfixed security vulnerabilities. The rest of this module uses v2, which does not contain the pkg/auth/docker package, so the subset of it that was used from here has been flattened into utils/oci.go, permitting the dropping of the dependency with the vulnerability.


dsymonds:
@mkmik

mkmik (approved):

